PR #5, opened 2016-03-28
Fixes for editing Authors and adds Typeahead for Author input

cervinko:
I had a lot of Authors 'Unbekannt' in my library. When i entered a Author in the Web interface all 'Unbekannt' where edited. With this fix a new Author is added to the db and linked to the book.
It would be better to use gettext instead of `if book.authors[0].name not in  ("Unknown", "Unbekannt", "", " "):` but that seemed to complex for me.

Added Typeahead for Author input

thats my first github pull request. i wanted to seperate it in two parts; didn't work.

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -723,7 +723,7 @@ def edit_book(book_id):
                 new_rating = db.Ratings(rating=int(to_save["rating"].strip()))
                 book.ratings[0] = new_rating
         db.session.commit()
-        if to_save["detail_view"]:
+        if "detail_view" in to_save:
             return redirect(url_for('show_book', id=book.id))
         else:
             return render_template('edit_book.html', book=book)
--- a/cps/web.py
+++ b/cps/web.py
@@ -665,11 +665,16 @@ def edit_book(book_id):
         to_save = request.form.to_dict()
         book.title = to_save["book_title"]
         
+        author_id = book.authors[0].id
+        
         is_author = db.session.query(db.Authors).filter(db.Authors.name.like('%' + to_save["author_name"].strip() + '%')).first()
         if book.authors[0].name not in  ("Unknown", "Unbekannt", "", " "):
             if is_author:
                 book.authors.append(is_author)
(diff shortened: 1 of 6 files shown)


janeczku:
Great PR, thanks! Will have a look at it soon!

janeczku:
Just had a quick look. Two problems came up:

1. Given the current author name is "Rolf Schmidt" and i correct the spelling to "Rolf Schmid" then the change is not  applied.
2. Given the current author name is "John Doe" and i correct the name to "Jane Doe"  then "John Doe" remains in the authors table even though there is no book associated with it.

cervinko:
Thank you for the feedback. Unfortunatelly i can not reproduce your the bugs. Did you look up the metadatachanges in the Authors List page? Because this one uses the 'sort' column of the authors table and this column is not edited by calibre-web at all. (but it clearly should be one day)

Otherwise 'Rolf Schmidt' should be changed to 'Rolf Schmid' for all books that had 'Rolf Schmidt' as author. The same with 'John Doe'. It changes the db entry of 'John Doe' but does not add a new entry. A new author entry is only added when the previous Authorname was 'Unbekannt', 'Unknown, ' '  or ''.

cps/web.py
```diff
@@ -655,7 +664,25 @@ def edit_book(book_id):
     if request.method == 'POST':
         to_save = request.form.to_dict()
         book.title = to_save["book_title"]
-        book.authors[0].name = to_save["author_name"]
+        
+        author_id = book.authors[0].id
+        
+        is_author = db.session.query(db.Authors).filter(db.Authors.name.like('%' + to_save["author_name"].strip() + '%')).first()
```
janeczku: Why are you doing a wildcard match of the author name?
The wildcard will match to_save["author_name"]="Rolf Schmid" with an existing author "Rolf Schmidt". Thats why the book author doesn't get changed in that case as i previously commented.
It should be an equal operator here.

cervinko: thx!. changed it to equal operator

janeczku:
Nice. Thanks a lot for your contribution!

PR #8, opened 2016-04-05
Upload function for PDF Files

cervinko:
Adds an upload button in the right nav-bar. Uploads file immediately after selection and then redirects to book-edit page. Flashes an Error if file != *.pdf

Also adds `def update_dir_stucture(book_id):` in the helper.py which renames the directories about the way calibre does.

Unfortunately i have change some tab indentations to spaces which makes this pull request quite messy. sorry.

Upload function for epub is in my todo, but not very important for me.

Would be nice to add an option to use calibredb command to add the uploads to the db. I think with calibredb there is less risk to mess up the db. but for myself dependency on calibredb is not an option.


janeczku:
Thanks. So this will allow adding PDF to existing books or create new books?
BTW, if you are using Sublime as editor it's really easy to fix tabs/spaces: http://stackoverflow.com/a/14773274

cervinko:
it will create new books only

janeczku:
Right, let me take a look.

cervinko:
I am using Geany. it has also a tab-space convert feature.
Should i change indentation back to the way it was?

janeczku:
No need. The current indentation is looking just fine in my editor. Since it's late i won't have time to test drive it until later tomorrow though.

janeczku:
```
 ~/code/calibre/python cps.py
  File "cps.py", line 9, in <module>
    from cps import web
  File "/Users/Janek/code/calibre-web/cps/web.py", line 24, in <module>
    from wand.image import Image
ImportError: No module named wand.image
```

Looks like this patch makes calibre-web dependent on ImageMagick. I really would like to keep the app self-sufficient without any installation requirements other than Python. Since the PDF is added to existing books i guess we can do without extracting the PDF cover?

cervinko:
no. it is not adding to existing books. it adds new books. 
but wand really depends on ImageMagick... perhaps i can find a solution with less dependencies.

cervinko:
ok. i think without imagemagick it will not really work. but as far as i know imagemagick is normally available on webhosting services.

how about adding an upload on/off option in the config?

janeczku:
Didn't think of that 😄   +

cps/web.py
```diff
@@ -21,6 +21,9 @@
 import base64
 from sqlalchemy.sql import *
 import json
+from wand.image import Image
```
janeczku: I think we can do without the extra config option.
Just catch the import error and set a variable and later use a generic cover if ImageMagick is not  installed:

```Python
try:
    from wand.image import Image
except ImportError, e:
    print "Could not import wand.image:"
    print e
    use_generic_pdf_cpver = True
```

cervinko:
i do not really know hot to include wand into our lib folder. do i just have to copy /usr/lib/python2.7/dist-packages/wand ??
Also I don't know if we should vendor it, because we can't know if it is compatile of the imagemagick installation... and "import wand" could pass although imagemagick is not installed.

janeczku:
Hi, wand is not linked to specific version of ImageMagick. Should be pretty flexible with various installations. See here: http://docs.wand-py.org/en/0.4.2/guide/install.html#explicitly-link-to-specific-imagemagick

> i do not really know hot to include wand into our lib folder. do i just have to copy /usr/lib/python2.7/dist-packages/wand ??

Yes, but remove all ".pyc" files

 > and "import wand" could pass although imagemagick is not installed.

True, `import wand` passes even without imagemagick, but `from wand.image import Image` actually checks whether imagemagick is available on import! 

Here is it how it looks on my Mac without ImageMagick:

```
 Janek@macmini  ~  python
Python 2.7.5 (default, Mar  9 2014, 22:15:05)
[GCC 4.2.1 Compatible Apple LLVM 5.0 (clang-500.0.68)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from wand.image import Image
  File "<stdin>", line 1, in <module>
  File "/Library/Python/2.7/site-packages/wand/image.py", line 20, in <module>
    from .api import MagickPixelPacket, libc, libmagick, library
  File "/Library/Python/2.7/site-packages/wand/api.py", line 205, in <module>
    'Try to install:\n  ' + msg)
ImportError: MagickWand shared library not found.
You probably had not installed ImageMagick library.
Try to install:
  http://docs.wand-py.org/en/latest/guide/install.html#install-imagemagick-on-mac
```

cervinko:
ok. i added the latest stable wand (4.2) to lib. seems to work fine.
i also checked after deinstalling libmagickwand from my system. without libmagickwand i get the generic cover and no errors.

janeczku:
Awesome!

cervinko:
TODO: create some fancy generic cover.jpg.

cervinko:
@janeczku is there anything that keeps you from pulling this?

janeczku:
@cervinko 
Did you edit your TODO comment? Last time i checked the second part wasn't there. That's what kept me from landing this 😄

cervinko:
thx!
i really edited it. but just a few seconds after submitting.

janeczku:
NP!

PR #13, opened 2016-04-15
Better Author and Tags editing, better multi-author support

cervinko:
lot of this is copied from https://bitbucket.org/SpeedProg/calibreserver

adds better typeahead for (multi)author, tags and series. also much better and cleaner code in the edit_book function.


janeczku:
Nice! I'll check it out tomorrow.

janeczku:
Just tried it out. Definitely a great improvement, thanks!

PR #15, opened 2016-04-20
fix for edit page after upload and messing around with config.ini

cervinko:


cervinko:
how can i tell git to ignore my changes to the config.ini??
i already have it in .gitignore and did git rm --cached config.ini

cervinko:
sorry for the mess with config.ini

janeczku:
What's the purpose of config.ini_example?

PR #16, opened 2016-04-20
Feature custom columns

cervinko:
adds support for calibre custom columns. the db schema for costum columns in calibre is really a mess.

'bool', 'datetime', 'int', 'comments', 'float' columns are currently ignored because the use a different sql schema and i don't need them....


janeczku:
Patch causes an exception every time one tries to save a book's metadata and there are certain custom column types in the DB.
I only tried the following three but there may be more:

- "Ratings" (datatype: rating)
- "Text with fixed set of permitted values" (datatype: enumeration)
- "Comma separated text like tags"  (datatype: text, is_multiple: yes)

We need to filter all column types for which we don't have an sqlalchemy interface that can write them back to the DB.

cps/db.py
```diff
@@ -49,6 +50,25 @@ def title_sort(title):
     Column('lang_code', Integer, ForeignKey('languages.id'), primary_key=True)
     )
 
+cc = conn.execute("SELECT id, datatype FROM custom_columns")
+cc_ids = []
+cc_exceptions = ['bool', 'datetime', 'int', 'comments', 'float', ]
```
janeczku: add 'rating' and 'enumeration'

cervinko: rating and enumeration works for me. i tested all types. could you post your metadata.db??

cps/templates/edit_book.html
```diff
@@ -43,6 +43,48 @@
...
+                    <option 
+                      {% if book['custom_column_' ~ c.id]|length > 0 %}
+                        {% if book['custom_column_' ~ c.id][0].value == opt %}selected="selected"{% endif %}
+                      {% endif %}
+                      >{{ opt }}</option>
+                  {% endfor %}
+              </select>
+            {% endif %}
+            
+            {% if c.datatype == ('rating' or 'float' or 'int') %}
```
janeczku: > 'bool', 'datetime', 'int', 'comments', 'float' columns are currently ignored
So why include them here?

janeczku:
Another issue: A three star rating in Calibre is saved in the DB as int of 6 (who knows why).
If you try to save this value in the web app it's blocked by JS:

screener

cervinko:
ok. i checked and my calibre version (2.5) saves the stars as stars*2. that means 2,4,6,8,10
could you check this again in your version. then i will fix it. although i think the input should still be 1-5. but i have so save the correct values.

janeczku:
Calibre v2.5.4 here. Stars are saved as stars*2.

cervinko:
fine. i will fix it tomorrow
i still can not reproduce your errors with  'rating' and 'enumeration'

janeczku:
It's working for me now. I will try to reproduce tomorrow. Thanks for working on this and g'd night!

janeczku:
Given a custom column of type "comma-separated text" with two entries there is a problem with populating the form field in the metadata editor:

textliketags

textliketags-editor

janeczku:
@cervinko Have you added config.ini to .gitignore? Because it popped up again with your settings in 	f57c9c5 😛 

> echo "config.ini" >> .gitignore

cps/templates/detail.html
```diff
@@ -58,6 +58,27 @@
...
+        <p>
+        <div class="custom_columns">
+        {% for c in cc %}
+          {% if entry['custom_column_' ~ c.id]|length > 0 %}
+            {{ c.name }}: 
+            {% for column in entry['custom_column_' ~ c.id] %}
+              {% if c.datatype == 'rating' %}
+                {{ '%d' % (column.value / 2) }}
+              {% else %}
+                {{ '%d' % (column.value / 2) }}
```
janeczku: Typo? This are all the other columns besides rating, so just strings:
`{{ column.value }}`

cervinko:
i am to silly for this config.ini thing.
is it ok if it gets deleted in the repo? i thought it should stay but the changes should not be tracked

cps/templates/edit_book.html
```diff
@@ -37,12 +37,47 @@
...
       </div>
       <div class="form-group">
         <label for="cover_url">Cover URL (jpg)</label>
         <input type="text" class="form-control" name="cover_url" id="cover_url" value="">
       </div>
+      {% if cc|length > 0 %}
+        {% for c in cc %}
+          <div class="form-group">
+            <label for="{{ 'custom_column_' ~ c.id }}">{{ c.name }}</label>
+            {% if c.datatype in ['text', 'series'] %}
```
janeczku: Replace this if block. That fixes the problem with the "Comma-delimited text" type. We also have to take special care of the columns with is_multiple==True in the save routine (treat like tags, split by comma, ...).

```Python
            {% if c.datatype in ['text', 'series'] and not c.is_multiple %}
              <input type="text" class="form-control" name="{{ 'custom_column_' ~ c.id }}" id="{{ 'custom_column_' ~ c.id }}"
              {% if book['custom_column_' ~ c.id]|length > 0 %}
                  value="{{ book['custom_column_' ~ c.id][0].value }}"
              {% endif %}>
            {% endif %}

            {% if c.datatype in ['text', 'series'] and c.is_multiple %}
              <input type="text" class="form-control" name="{{ 'custom_column_' ~ c.id }}" id="{{ 'custom_column_' ~ c.id }}"
              {% if book['custom_column_' ~ c.id]|length > 0 %}
              value="{% for column in book['custom_column_' ~ c.id] %}{{ column.value.strip() }}{% if not loop.last %}, {% endif %}{% endfor %}"{% endif %}>
            {% endif %}

```

janeczku:
Do you want the config.ini_example to go in the repo or was that just for your own purposes?

janeczku:
> is it ok if it gets deleted in the repo?

Just overwrite the one in your branch with the config.ini from master and add to the commit once we are ready to merge this.

janeczku:
We are almost there. Just have to fix https://github.com/janeczku/calibre-web/pull/16#discussion_r60573954 and we can merge this.

cervinko:
so i have to change the config.ini manually before every public commit?

janeczku:
No, you just have to add it to the .gitignore (see above).

cervinko:
but then i get

> error: Your local changes to the following files would be overwritten by checkout:
        config.ini
Please, commit your changes or stash them before you can switch branches.
Aborting
1 christoph@

> 
when i want to checkout a different branch

janeczku:
[Gitter](https://gitter.im/janeczku/calibre-web?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)

janeczku:
I think that's because you edited config.ini before adding it to gitignore so it's already tracked by git.
Try `git checkout config.ini` and then switch branch

PR #38, opened 2016-07-18
Update Google Font protocol to HTTPS

jgillman:
Google Font urls now default to HTTPS which is good practice, plus it will fix the below error for servers using SSL.

screen shot 2016-07-18 at 11 13 00 am


cervinko:
thx!

Issue #28, opened 2016-06-18
remove menu from login page


perler:
maybe I am missing something, but as the menu (hot/new/discover) doesn't work before login, I think it should be removed from the login page..

Issue #23, opened 2016-06-11
make "Discover (Random Books)" optional


perler:
this maybe a good category if you sell books but for my personal library it is quite unnecessary, so I think especially in the "new books" area this should at least be optional?

cervinko:
i think discover should be removed from "new books"

PR #41, opened 2016-07-24
Fix for #33, #28, #23

OzzieIsaacs:


OzzieIsaacs:
Please don't merge this. I introduced some more errors which need to be fixed before.

PR #87, opened 2017-01-07
Use pip for dependency installation

akhy:
Because committing vendor libraries to Git is a bad practice.


OzzieIsaacs:
@janeczku what do you think about this? Shall we remove the vendor folder and leave it to the users to get the needed vendor files?

janeczku:
👍  for switching to requirements.txt

akhy:
created `requirements.txt` by guessing from vendor contents, still unsure if I missed a thing... 
I'm running an instance using Docker image built from this branch and still found no problems though

PR #88, opened 2017-01-07
Allow option overriding via env vars

akhy:


OzzieIsaacs:
I'm thinking of getting rid of the config file and putting every setting to the sqlite database. So currently I'm not sure if this PR is going into the right direction. 

I need to think a while on it.

OzzieIsaacs:
Meanwhile the whole configuration is done via the web-ui. I don't see a benefit from overriding the config via ENV-variables. What do you think?

PR #92, opened 2017-01-17
fixed wrong translations with Simplified Chinese

idalin:
fixed wrong translations with Simplified Chinese.
added a new identifier Douban(mostly for Chinese books.).
fixed smtp with ssl has no response.


OzzieIsaacs:
So far so good. Regarding the SSL connection for email sending we have to modify it. There are two mechanisms:

1: The old implementation with STARTTLS (usually port 587) -> most german provider use this
2: The new implementation with SSL/TLS (usually port 465) -> gmail uses this and your provider ;-)

We have to change the SSL option to a single select drop down menu to switch between both of them. I can do this probalby tommorrow evening and then I can merge your PR. 

Thanks a lot for pointing to this weakness in the code

Issue #94, opened 2017-01-19
Send to Kindle - issue?


bodybybuddha:
Hello - just downloaded and installed the repo today.  AMAZING!!

Having multiple issues - but that maybe b/c I haven't fully explored the site enough yet.  

However, I did setup a profile with a kindle email address, save the settings and sent email test.  Kindle sent back an email saying i didn't have a file attached (sorta expected.)  However, when sending a book via the email option, no book ever arrives on my Kindle.

I've verified allowed email addresses, I've sent a mobi file via an email client with no issue, etc.  

Notice that there was a change to the repo about 8 hours ago as of this writing, would that have caused the issue?

Thanks!

OzzieIsaacs:
The change was related to encryption on login to send emails. I take a look.

simkin:
Working correctly on:

> Program library	Installed Version
> Python	2.7.9 (default, Jun 29 2016, 13:08:31) [GCC 4.9.2]
> Kindlegen	Amazon kindlegen(Linux) V2.9 build 1028-0897292
> ImageMagick	ImageMagick 6.8.9-9 Q16 x86_64 2016-11-26 http://www.imagemagick.org
> PyPDF2	1.26.0

bodybybuddha:
I'm slightly off on the versions:

Python	2.7.6 (default, Oct 26 2016, 20:30:19) [GCC 4.8.4]
Kindlegen	Amazon kindlegen(Linux) V2.9 build 1028-0897292
ImageMagick	ImageMagick 6.7.7-10 2016-11-29 Q16 http://www.imagemagick.org
PyPDF2	1.26.0

I'll update a little later today and see if it makes a difference!

Thanks!

bodybybuddha:
I've set the Development flag and restarted.  I haven't upgraded any of the libraries yet (working on that next.)  

In any event attached is the entries from the log file (removed any private information and auth code, etc.) and i'm still experiencing the same issue.

[calibre-web-log.txt](https://github.com/janeczku/calibre-web/files/717900/calibre-web-log.txt)
 
The website gives me the green bar and states success - but the above is the entire log file.

I'll report back  after I get the latest and greatest software.

bodybybuddha:
Almost there:

Python	2.7.12 (default, Jul 18 2016, 15:02:52) [GCC 4.8.4]
Kindlegen	Amazon kindlegen(Linux) V2.9 build 1028-0897292
[...]

PR #113, opened 2017-02-15
Limit the extensions of uploading files

idalin:
Limit the extensions of uploading files. 
These file extensions are allowed:
'txt', 'pdf', 'epub', 'mobi', 'azw', 'azw3', 'cbr', 'cbz', 'cbt', 'djvu', 'prc', 'doc', 'docx', 'fb2'


cps/web.py
```diff
@@ -2115,6 +2116,13 @@ def upload():
     db.session.connection().connection.connection.create_function('uuid4', 0, lambda: str(uuid4()))
     if request.method == 'POST' and 'btn-upload' in request.files:
         file = request.files['btn-upload']
+        if not ('.' in file.filename and file.filename.rsplit('.', 1)[1].lower() in ALLOWED_EXTENSIONS):
+            flash(
+                _('File extension "%s" is not allowed to be uploaded to this server' % 
+                file.filename.rsplit('.', 1)[1].lower()),
```
janeczku: This will cause a `list index out of range` error when user uploads a file without any dot, i.e without an extension

cps/web.py
```diff
@@ -52,6 +52,7 @@
 
 # Global variables
 global_task = None
+ALLOWED_EXTENSIONS = set(['txt', 'pdf', 'epub', 'mobi', 'azw', 'azw3', 'cbr', 'cbz', 'cbt', 'djvu', 'prc', 'doc', 'docx'])
```
janeczku: fb2 format seems popular as well

idalin:
fixed send to kindle not working problem for some smtp provider. #94

PR #118, opened 2017-02-19
Now shows read in website, and can be toggled by clicking, two extra …

JackED42:
…options in sidebar (read/unread books)


JackED42:
Closed due to adding extra features

PR #121, opened 2017-02-22
Google Drive for hosting books + Mark books as read

JackED42:
Hi,

I created a previous pull request, but I think this build is better, as it contains a full google drive implementation, including authentication etc. It still needs extra work for syncing metadata.db automatically, it currently has the feature, and will update if the file is changed (e.g. via calibre itself), but it has to be manually refreshed once a week. 

I also need to add a README for setting up google drive API for your account, and getting client details etc.

Feel free to ask me any further questions, or just whether or not you think this is a good idea for the main version.

EDIT:

I am planning on getting your latest commits from master and putting them on to my master

Cheers


JackED42:
I have just updated the README with directions for google drive integration. 

Also, thanks for deleting the old pull request

JackED42:
Hi, I was wondering if you have any plan of integrating either the change for read/unread or google drive in to the main branch? If so, I would be happy to clean up my branches, and separate them out individually.

OzzieIsaacs:
Yes I'm planing to do it. But currently I'm still stuck with #87 and the problem with the non updating database connection. This is the next think I want to solve, than I can take a look at this one.
I'm a bit busy too, currently, so I only merge the easy pull requests.

JackED42:
Great good to know. Will try and keep up to date with master then to keep it easy for you! If there is anything I could help with just let me know...

Cheers

JackED42:
Another question.... What is the reason for using Tornado as a HTTP server, as apposed to flask app run?

I have just changed to use flasks built in web server, and have noticed that it seems to be a lot faster. It also gives me the ability to stream the downloads of the files from google drive directly, which means that I do not have to wait for the download to finish before seeing it is being downloaded.

I haven't committed those changes yet onto the develop branch, because I wanted to ensure there was no specific reason this is being used. After some googling, it appears you can't enable flask/tornado combination to allow streaming. 

The only reason I have enabled the streaming option, is because when using an OPDS browser on my kobo, it times out relatively quickly, and by changing to this option, I no longer have this problem.

OzzieIsaacs:
Well, the documentation says, don't do it:
http://flask.pocoo.org/docs/0.12/deploying/

JackED42:
Thanks for the pointer. I have added the ability to use gevent (recommened in the documentation). 

It means I can now successfully use yield for large files

It is optional, using -g

OzzieIsaacs:
Hello @JackED42 ,

thanks for all the effort you put into this pull request. It looks very good to me. One thing I'm not happy about is the new http-server and the dependencies it introduces. Could you please make these part of the code optinonal (try: import... except...), so not everybody needs the libs. Another problem with the new sever is that it breaks the autoupdater (in config search for updates) could you please also take care of this.

Thanks a lot

JackED42:
I will take a look now. Leave it with me

JackED42:
Sorry, I am misunderstanding 

> the class form-control in config use_google drive make the checkbox very bit, could you please remove it

Can you please clarify?

JackED42:
screen shot 2017-03-12 at 20 59 28

Is this better?

JackED42:
And for detail.html, I will make it a checkbox

JackED42:
I am really struggling with the updater with gevent. Have you got any ideas?

JackED42:
Never mind, I think I have got it working.

Just let me know if you need further changes.

Thanks

OzzieIsaacs:
The checkbox looks very good on the screenshot, I'll take a look at the changes (hopefully) tomorrow.

Kennyl:
Maybe use bootstrap toggle button instead of checkbox

JackED42:
I just realised what happened.... When I did the update it replaced my changes, so I didn't commit the all. Anyway, I think I have a fix now with the latest changes. IF you could check that would be great. 

I think the updater needs to be changed to not replace the .git folder! Especially for local development, I think it should keep the .git folder as it.

I have decided to keep the update_dir_structure for gdrive in the helper class, I think it makes more sense there, just so it is inline with the "non gdrive version".

Cheers

JackED42:
Oh, just remembered you wanted gdrive optional too. Will work on that and update when finished

JackED42:
Gdrive should be optional now too

JackED42:
Hey @OzzieIsaacs, any chance to take a look at these latest changes?

Kennyl:
@JackED42 can you specify exactly pydrive and google-api-python-client version,
I try many combination but didn't work.

Kennyl:
@JackED42 sorry it was misconfiguration on my MACOS.

Kennyl:
Download not work after this PR, have you modified something in cookie ?

'''
 File "/Library/Python/2.7/site-packages/werkzeug/wrappers.py", line 1014, in set_cookie
    self.headers.add('Set-Cookie', dump_cookie(key, value, max_age,
AttributeError: 'dict' object has no attribute 'add'
'''

cps/web.py
```diff
@@ -689,21 +789,55 @@ def feed_series(id):
...
+    app.logger.info (time.time()-startTime)
+    startTime=time.time()
+    if config.config_use_google_drive:
+        df=gdriveutils.getFileFromEbooksFolder(Gdrive.Instance().drive, book.path, data.name + "." + format)
+        return do_gdrive_download(df, headers)
+    else:
+        # file_name = helper.get_valid_filename(file_name)
+        response = make_response(send_from_directory(os.path.join(config.config_calibre_dir, book.path), data.name + "." + format))
+    response.headers=headers
     return response
```
Kennyl: response.headers is a class of wsgiref.headers.Headers or  werkzeug.datastructures.Headers() , not dict()
Please download logic without gdrive will never work.

cps/web.py
```diff
@@ -1326,13 +1651,18 @@ def get_download_link(book_id, format):
...
             pass
-        response.headers["Content-Disposition"] = "attachment; filename*=UTF-8''%s.%s" % (quote(file_name.encode('utf-8')), format)
+        headers["Content-Disposition"] = "attachment; filename*=UTF-8''%s.%s" % (urllib.quote(file_name.encode('utf-8')), format)
+        if config.config_use_google_drive:
+            df=gdriveutils.getFileFromEbooksFolder(Gdrive.Instance().drive, book.path, '%s.%s' % (data.name, format))
+            return do_gdrive_download(df, headers)
+        else:
+            response = make_response(send_from_directory(os.path.join(config.config_calibre_dir, book.path), data.name + "." + format))
+        response.headers=headers
         return response
```
Kennyl: As previous comment.

Kennyl:
Please fix headers={} in web.py @JackED42

Issue #122, opened 2017-02-23
Error getting metadata from certain Epubs


JackED42:
Hi, 

I just noticed that when I was uploading a few books, no metadata was populated and no image.

I just tested via command line, and noticed that the problem seems to be, that some epubs store the cover.jpg in the root (not just in OPS).

I managed to fix it locally, by just changing epub.py extract cover to: -

```
def extractCover(zip, coverFile, tmp_file_name):
    if coverFile is None:
        return None
    else:
        try:
            cf = zip.read("OPS/" + coverFile)
        except KeyError, e:
            cf = zip.read(coverFile)
        prefix = os.path.splitext(tmp_file_name)[0]
        tmp_cover_name = prefix + "." + coverFile
        image = open(tmp_cover_name, 'wb')
        image.write(cf)
        image.close()
        return tmp_cover_name

```

so that it checks both.

PR #124, opened 2017-02-23
Fixed epub upload metadata reading 

JackED42:
I raised an issue (#122), but thought I would just rebase my master and create a pull request for you (I am a bit new to using GitHub, trying to learn the ropes). 

Basically, you currently search for the cover.jpg in the OPS folder, but that doesn't seem to be the location of the covers for a few new books I tried to upload, then are just in the root.


OzzieIsaacs:
I solved it a more generic way, reading the correct filepath from zipfile

PR #128, opened 2017-02-28
[New Feature]Get metadata from Google and Douban while editing book.

idalin:
This PR bring a new feature.
It search book both in Google Books and Douban Books(better for Chinese books). Show the results. And load the metadata by clicking the cover.
No i18n.


OzzieIsaacs:
Please make the UI translatable, and please make the texts english in default.
I'm not going to merge it without translation.
Thanks

idalin:
Ok. It's English in default indeed.
I didn't want to import i18n js files, so it's not translatable for the js file. I will make it translatable in babel way in a few days.

idalin:
Done! Texts in get_meta.js are moved to book_edit.html  and translatable.

PR #134, opened 2017-03-02
Update epub.py

Kennyl:
initialized covefile to None first
another option add `else: coverfile = None` clause at 74:75


cps/epub.py
```diff
@@ -71,8 +72,6 @@ def get_epub_info(tmp_file_path, original_file_name, original_file_extension):
                 meta_cover_content = tree.xpath("/pkg:package/pkg:manifest/pkg:item[@id='"+meta_cover[0]+"']/@href",namespaces=ns)
                 if len(meta_cover_content) > 0:
                     coverfile = extractCover(zip, meta_cover_content[0], coverpath, tmp_file_path)
```
Kennyl: another solution
```
                    else:
                        coverfile = None
```

PR #141, opened 2017-03-05
Fix attachment filename

Kennyl:
File attachment should be url encode rather than pure utf-8 format when unidecode unavailable

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -673,9 +673,9 @@ def get_opds_download_link(book_id, format):
     file_name = book.title
     if len(book.authors) > 0:
         file_name = book.authors[0].name + '-' + file_name
-    # file_name = helper.get_valid_filename(file_name)
+    file_name = helper.get_valid_filename(file_name)
     response = make_response(send_from_directory(os.path.join(config.config_calibre_dir, book.path), data.name + "." + format))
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Could you please give me an example of a non working filename, so I can include it in my secret (to be published) testing scripts.

Kennyl:
https://www.w3.org/Protocols/rfc1341/7_3_Message.html
`Non-ASCII text in the headers of an encapsulated message can be specified using the mechanisms described in [RFC-1342].`

I only find similar answer, but not exactly.
http://stackoverflow.com/questions/33914422/how-can-i-get-effect-of-html-into-emails-subject-line

Kennyl:
or just refer logic to route
/opds/download/
no sanitization  using data.name instead of concentrating title+ author

cps/web.py
```diff
@@ -673,9 +673,9 @@ def get_opds_download_link(book_id, format):
     file_name = book.title
     if len(book.authors) > 0:
         file_name = book.authors[0].name + '-' + file_name
-    # file_name = helper.get_valid_filename(file_name)
```
Kennyl: '-' with remove at get_valid_filename

Kennyl:
you may try 
日経新聞電子版(Free, MAX).recipe  
in calibre fetch news as example

PR #142, opened 2017-03-05
[Fix] Download Name format

Kennyl:
Expected
author-title.ext
But now "-" is sanitized at helper class.
As a Result,
authortitle.ext
Now fix it with following format
author_title.ext


OzzieIsaacs:
Fixed it by repairing the helper class

PR #152, opened 2017-03-08
[Python3] Remove dependency on future function

Kennyl:
no need six or future if there is only one function.
Part of #143


OzzieIsaacs:
@Kennyl thanks for all the contributions and great work, it pushs the projects to a new level.

PR #153, opened 2017-03-08
Fix #146

### LINKED ISSUE #146: Kobo Issue - Downloading books
Hi, 

If you try and download books with the built in browser on Kobo, you can only download books, if the extension is .epub/.mob etc. It cannot be a link that doesn't end in a file, e.g. book/1/download.

If you want to support kobo, there are 2 ways of fixing this: -

1. Make the download links end in the file name
2. Have a separate download for Kobo, that ends in the file name. 

I don't mind doing a pull request and making the change, but would just like your opinion on which option you think would be best.

Kennyl:
#146 
for web gui only
opds remains same behaviour

### DIFF
--- a/cps/templates/detail.html
+++ b/cps/templates/detail.html
@@ -126,7 +126,7 @@
               </button>
               <ul class="dropdown-menu" aria-labelledby="btnGroupDrop1">
                 {% for format in entry.data %}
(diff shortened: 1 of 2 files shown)


JackED42:
Is it worth making the epub name "title - author.epub" or "author - title.epub"?

Kennyl:
The name is meaningless, since the real filename was determined by HttpResponse Multipart.
And I think a filename appear in url, is beyond RESTFUL structure already (those should appears in payload only).

PR #154, opened 2017-03-09
Fix #146

### LINKED ISSUE #146: Kobo Issue - Downloading books

Kennyl:
For WebUI only
OPDS remains previous behaviour
using   /download/<book_id>/<book_id>.< format>
Fix #153 PR introduce bug

### DIFF
--- a/cps/templates/detail.html
+++ b/cps/templates/detail.html
@@ -126,7 +126,7 @@
               </button>
               <ul class="dropdown-menu" aria-labelledby="btnGroupDrop1">
                 {% for format in entry.data %}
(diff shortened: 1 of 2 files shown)


JackED42:
Hi, sorry for delayed response. The kobo actually picks the filename from the name in the url as apposed to in the headers (really weird). book_id.format willl probably be ok though I guess

PR #164, opened 2017-03-23
fix the utf8 file name save to db

ferocknew:


Kennyl:
I think  'SET LC_LOCALE="UTF-8' is better.

Kennyl:
```
reload(sys)
sys.setdefaultencoding('utf-8')
```
are discouraged, stackoverflow contains  lots of context about this question.

If you find any unicode errors, please report again, those findings affect Python2/3 compatibility.

OzzieIsaacs:
I agree with @kennyl.
Could you please tell us, what the exact problem is? Which path/file didn't work?

OzzieIsaacs:
no response

PR #177, opened 2017-04-10
Download Style Change

### LINKED ISSUE #40: layout of book page makes download on Kindle hard
the single book page (with the download buttons below the cover) is almost unusable on the kindle. to download a book the dropdown for epub/kindle opens and the kindle entry is very close to the bottom of the page, so you need a lot of trial and error before you hit the entry. I think the screen on the kindle is not very sensitive so close to the frame. 

40 pixel margin-bottom should solve this problem..

Kennyl:
- No more dropdown button
- Add icon to download icon image are from  https://www.iconfinder.com/iconsets/document-file
- Fix #40


Kennyl:
or another candidate iconset https://www.iconfinder.com/iconsets/document-file

OzzieIsaacs:
@Kennyl 
Thanks for the work. For using the icons please follow the rules on https://creativecommons.org/licenses/by/3.0/ ( I suggest including it on the about page)
I'd prefer the second suggested icons.
Having separate buttons for different eBooks formats looks like a very good idea to me. (Specially on eBook-reader)

Kennyl:
Now added license on /stat/ and using https://www.iconfinder.com/iconsets/document-file

Kennyl:
Btw or simply no icon, just use font-awesome font-icon http://fontawesome.io/icon/file/ .

Kennyl:
More Example and wait for comment

screen shot 2017-04-12 at 9 18 58 pm

gabriellluz:
One problem: when using Kindle Paperwhite (first model), the link text in green (e.g. EPUB, AZW) is barely noticeable. I suggest you change it to white.

OzzieIsaacs:
I had a chance to take deeper look at the changes you proposed. It appreciate all the work you put into this. The reason it took so long was, because I wasn't 100% happy with the look in case of having many books so i decided to leave it as is before for books having more than 3 book formats to download. Furthermore I think the shadow effcet looked awesome, but is a desing break compared the rest of the user interface, so I decided to remove it (on ebook reader I think without shadow its easier to read). Last point, I decided to not take the icons, because they also look to diffrent from the rest of the icons (the font-awesome file-icon fits better from my point of view

perler:
very nice so far but green on blue?! ;)

OzzieIsaacs:
grafik
For me its working?

perler:
nope

OzzieIsaacs:
Looks like caching issue. Could you please restart your browser, empty cache....

perler:
yep, sorry, my mistake, everything is fine.

btw, for the icons maybe http://fontawesome.io/icon/book/ and http://fontawesome.io/icon/amazon/ ?

PR #181, opened 2017-04-16
Add a check for metadata.db file.

issmirnov:
Without this check, the app tries to invoke sql alchemy on a non existing file and prints a confusing stack trace. This commit adds an explicit check for existence of the database, and quites the app with a non zero exit code if the check fails.


OzzieIsaacs:
Your fix creates the problem that once entered wrong, you never can recover from this situation. Icould reproduce the problem and fix it.

PR #208, opened 2017-06-22
Add TinyMCE to provide rich-text editing for descriptions

jkrehm:
I'm not sure if it's new with Calibre 3 (I've not been a heavy user in the past) but Calibre's description editor has a rich-text editor and in my (limited) experience the descriptions often contain HTML.

This PR adds support for TinyMCE to make calibre-web's editor rich-text enabled.

If previous versions of Calibre were not rich-text then it might make sense to have this as a configuration option. If that's the case, let me know and I'll try to add that in.

screen shot 2017-06-22 at 9 00 43 am


OzzieIsaacs:
Calibre 2 has also support for rich-text descriptions. Thus a configuration option is not needed. I'll check the PR and merge it afterwards.

jkrehm:
Thanks!

Issue #109, opened 2017-02-12
Add 'All Books' section


brianbuquoi:
Add 'All Books' section for alphabetical browsing.

OzzieIsaacs:
Fixed with newest commit

Issue #108, opened 2017-02-11
Add "Recently Published" category


kriegschrei:
Request to add a category for recently published.  "Newest" looks like it is actually "Recently Added".  It would be nice to see the most recently released books in collection.

OzzieIsaacs:
Fixed with newest commit

PR #218, opened 2017-07-06
Add sort options when viewing books

jkrehm:
Default view still shows the latest books, but includes ability to show oldest books, and books sorted alphabetically (ascending & descending).

I did not include translations for the next text, but they are split up for easy translating and are otherwise ready for translating.

screen shot 2017-07-06 at 9 31 10 pm

Please let me know if there's any configuration that you would like me to include or any feedback you have.


jkrehm:
Note: Would resolve #109

jkrehm:
#108 could probably be merge into this. Perhaps keep the "Newest Books" link (rename it to "Recently Added") and use "Newest Books" and "Oldest Books" in the dropdown to sort by published date. Thoughts?

jkrehm:
Anything I can do to improve this PR?

PR #221, opened 2017-07-06
Show book details in modal dialog

jkrehm:
With them shown in a modal, you don't lose your place in the pagination. If the request comes via Ajax, the minimal layout is used. If via a normal request, the full layout is used. That lets you open the details in a new tab and have the full experience, but if you're clicking through the results of a search, you can view many without losing your place.

While I was going through my library, marking everything that I'd read, I was annoyed by the need to open the books in a new tab or risk losing my place in the pagination.

Please let me know what feedback you have. Thanks!

screen shot 2017-07-06 at 9 25 25 am

### DIFF
--- a/cps/static/js/main.js
+++ b/cps/static/js/main.js
@@ -121,6 +121,17 @@ $(function() {
         });
     });
 
+    $('#bookDetailsModal')
+      .on('show.bs.modal', function(e) {
+        var $modalBody = $(this).find('.modal-body');
+        $.get(e.relatedTarget.href).done(function(content) {
+          $modalBody.html(content);
+        });
+      })
+      .on('hidden.bs.modal', function() {
+        $(this).find('.modal-body').html('...');
+      });
+
     $(window).resize(function(event) {
         $(".discover .row").isotope("reLayout");
     });
--- a/cps/static/js/main.js
+++ b/cps/static/js/main.js
@@ -121,15 +121,15 @@ $(function() {
         });
     });
 
-    $('#bookDetailsModal')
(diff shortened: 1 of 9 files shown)


Kyosfonica:
I get this instead of the expected modal:

image

jkrehm:
That is the modal, but it's not been populated (the modal is populated by
am ajax request). Are there JS errors in the console when you get that? And
what does the network tab show?

Jonathan Rehm

Kyosfonica:
Indeed was a JS issue. I restarted the app and now it works fine.

Kyosfonica:
This commit is bugged. The book opens in a modal view but the page loaded is not just the book page but everything, including header and sidemenu.

image

jkrehm:
You restarted the service after pulling the latest? Because there are changes to `web.py` and `detail.html` to only show the layout (the wrapper) if not `request.is_xhr`.

Kyosfonica:
Yes, I've restarted several times now.

jkrehm:
Hmm, strange. Can you confirm that at the top of `templates/detail.html` you have `{% extends is_xhr|yesno("fragment.html", "layout.html") %}`?

Kyosfonica:
It was a mix of a bad fork (I didn't have that change) and docker not updating the repo correctly.
I managed to sort it out.

Thanks,

jkrehm:
I'm glad you got it figured out. Hopefully others don't run into the same.

Jonathan Rehm

larabr:
I can't seem to find how to open the book details in a separate tab. Was the possibility removed altogether?

PR #225, opened 2017-07-08
Add "magic link" functionality

jkrehm:
When using a device that is bothersome to log in on (e.g. a Kindle) you can use a magic link to log in via another device.

Configuration was added and is disabled by default.

I recognize this is very firmly in the enhancement category, and you might not be interested. But when I went to log in on my Kindle and looked at my 30 character password I thought "there has got to be a better way". In case I'm not alone in that pain, I figured I'd chance a pull request.

screen shot 2017-07-07 at 6 32 44 pm

screen shot 2017-07-07 at 6 33 01 pm

### DIFF
--- a/cps/redirect.py
+++ b/cps/redirect.py
@@ -1,0 +1,25 @@
+# http://flask.pocoo.org/snippets/62/
+
+from urlparse import urlparse, urljoin
+from flask import request, url_for, redirect
+
+
+def is_safe_url(target):
+    ref_url = urlparse(request.host_url)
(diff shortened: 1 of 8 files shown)


OzzieIsaacs:
I also know how easy it is to log in on a reader.
Its not super ideal, because you have to enter the link which might be as long as the password, but I don't have a better solution.

jkrehm:
The full link might be as long (with typing the URL) but at least you have the option of doing it with a keyboard where you can type easily.

Thanks for merging and hopefully others can iterate on the design to improve it.

PR #227, opened 2017-07-09
Add ability to edit publish date

jkrehm:
Add ability to edit publish date. If a date is not set, default to "0101-01-01".

Uses HTML5 input "date" type. Chrome provides a UI for selecting dates. Firefox does not, and I imagine many other browsers do not. Ref http://html5doctor.com/the-woes-of-date-input/#feature-datepicker

Should I include a datepicker library? If so, any preference? jQuery-UI's is probably the most popular, but it's been a while since I've looked at datepickers, so there might be a preferred alternative.


OzzieIsaacs:
I recently used bootstrap-datepicker (https://github.com/uxsolutions/bootstrap-datepicker). It brings translated UI and needs minimal js

jkrehm:
Sounds good. Browsers that don't have their own datepicker built in (I'm using Modernizr to check that) will use bootstrap-datepicker. That means Chrome (desktop & Android), Opera, Edge, and iOS (Safari) will use their built-in datepicker. Firefox, Safari (desktop), and IE will use bootstrap-datepicker.

OzzieIsaacs:
Trouble: The display of the date should be in localized format. Can be easily archied by removing the format line in:
    ```$('#pubdate').datepicker({
      language: '{{ g.user.locale }}'
    });```
But than the return value is also in local format....

jkrehm:
The [HTML5 date input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) needs the date to be in yyyy-mm-dd format to work, so that's what I've gone with across all browsers, regardless of whether bootstrap-datepicker is used or not. I'm going on the assumption that eventually all browsers will build datepickers in for `<input type="date">`.

I agree that displaying the date should be localized, but for editing it, I sticking with ISO-8601 format (needed for HTML5 dates) will simplify things greatly. Otherwise, we need to transform the date immediately before POST'ing it.

jkrehm:
I just pushed a commit to show a "fake" input over top of the real input when bootstrap-datepicker is in use. The fake input listens for changes to the real input and shows the date in localized format. Browsers that support `<input type="date">` natively are not affected and will show the date string in the format their browser/OS is configured to use.

As part of the change I moved the datepicker code into `edit_books.js` and then moved the TinyMCE code, too. That way it's all contained within `edit_books.js`.

jkrehm:
Rebased onto `master` to resolve conflicts. Part of 5715d27 (a style change) already got merged into `master`, likely while resolving conflicts.

PR #228, opened 2017-07-10
[Fix] Associate label with correct input

jkrehm:
Same fix made with 92f634b6a6a2bcabaa4263ba4ec4e353f1b4bc26, different template.


OzzieIsaacs:
Found and fixed it this morning

jkrehm:
Is the fix pushed to a different branch? Because I see the fix to `config_edit.html` but not the one to `user_edit.html`.

OzzieIsaacs:
Sorry didn't read it right

PR #230, opened 2017-07-11
Send mail from the FROM address instead of the login

jkrehm:
When SMTP is configured to use a service that doesn't use an email address to log in – e.g. AWS SES or Postmark – the sendmail request fails saying the FROM address is invalid.

With this change, I was able to send using both Gmail and AWS SES. Previously, AWS SES failed saying "Invalid MAIL FROM address provided" because the IAM user access key is not an email address.


jkrehm:
Thanks!

PR #245, opened 2017-07-22
Use ajax to add/remove books from shelves

jkrehm:
Instead of redirecting the user away from the page or causing a full refresh, the user is able to add books to multiple shelves in quick succession.

If JavaScript is disabled, it gracefully falls back to a standard request (the current behavior).

shelf-actions


jkrehm:
Thanks!

PR #246, opened 2017-07-23
Do not attempt to create the YAML file when Google Drive is disabled

jkrehm:
Ran into this while testing an unrelated change a couple weeks ago.


jkrehm:
Thanks @OzzieIsaacs

PR #255, opened 2017-07-29
Add eslint file

### LINKED ISSUE #242: [Question] Codacy standards
Are the Codacy standards documented anywhere? I couldn't find anything on their site and there isn't an eslint file in the project. So right now it's a matter of "send a PR, hope for the best". I would love to help make it less of a gamble.

jkrehm:
Attempts to pattern what has been in use by Codacy. Currently, it eliminates support for IE10 and below, and some aspects of our code eliminate IE11, leaving Edge as the only supported MS browser. I don't think this is intentional.

Assuming we want to continue supporting at least IE 11, we should (a) change all of the code to EMC5 standards or (b) use [Babel](https://babeljs.io/) to transpile to EMC5.

Thoughts?

Resolves #242

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+
+[*.{js,py}]
+indent_size = 4
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,95 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true,
+    "jquery": true
+  },
+  "globals": {
+    "alert": true
+  },
+  "rules": {
+    "arrow-parens": 2,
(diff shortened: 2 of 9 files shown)


OzzieIsaacs:
I support solution (a). And yes eliminating support for IE11 was unintentionally

(I only took the standard settings from codacy and didn'T check it's implications)

jkrehm:
Sounds good. I'll update the .eslintrc file and then the JavaScript to match the new requirements.

How far back are you interested in supporting? Is IE 11 good enough?

jkrehm:
Updated to ES5 standards. Not sure how far back things will work, but it should definitely go back to IE 11 now.

jkrehm:
Awesome, thanks @OzzieIsaacs. Hopefully that resolves the IE 11 issues. I realized this morning that iOS has been having issues, too, because of the ES6 syntax.

PR #268, opened 2017-08-08
Add option to hide mature content from some users

### LINKED ISSUE #262: Hide books from user
Is there a way to hide certain books from some user's?

jkrehm:
Uses a configurable list of tags to denote what is mature and hides those books from the user.

screen shot 2017-08-08 at 10 41 08

screen shot 2017-08-08 at 10 40 54

All current users will have "Show mature content" checked and will have to manually be changed after updating to this. New users will have it unchecked by default. I figure that it is better to be conservative with new users while maintaining the same functionality current users have been experiencing.

Resolves #262.

### DIFF
--- a/cps/templates/user_edit.html
+++ b/cps/templates/user_edit.html
@@ -40,45 +40,48 @@
             {% endfor %}
         </select>
     </div>
-      <div class="col-sm-6">
-
-    <div class="form-group">
-        <input type="checkbox" name="show_random" id="show_random" {% if content.show_random_books() %}checked{% endif %}>
-        <label for="show_random">{{_('Show random books')}}</label>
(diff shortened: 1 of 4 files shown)


cps/templates/user_edit.html
```diff
@@ -40,45 +40,48 @@
             {% endfor %}
         </select>
     </div>
-      <div class="col-sm-6">
```
jkrehm: FYI, the large number of changes here is just re-indenting the HTML to standardize it.

jkrehm:
Apparently there's an issue related to this (reported via #276). I'll take a look tomorrow; it shouldn't be a complicated fix.

Issue #219, opened 2017-07-06
[Enhancement] Option to merge books of the same name and author


mercurycoolman:
If possible could an option be added to merge books of the same name and author for different filetypes. I have a .mobi of a book that shows no cover but i want to merge with a .pdf and have them all show on the same book page so i can select which file type i would like.

Thanks for any input in advance.

Kyosfonica:
The functionality of handling differents formats of the same book is there. What should be implemented are the form and the backend responsible of uploading a new format of a book and put that info in the database

What I wouldn't do is to merge uploaded books into one single entry. The proper way to do this is to update an already uploaded book and add the new file to it. This way prevents unwanted duplicated book entries in the database.

I will give it a try in my spare time.

This is what you would like to see, isn't it?
![image](https://user-images.githubusercontent.com/5280026/28833104-b2421e5e-76de-11e7-97ca-56dd08f66477.png)

OzzieIsaacs:
Finally...
There is a new option in the navbar, where you can (manually) select books and merge them

dsm1212:
I'm a little confused on mulit-format, lets say I have a book with 2 formats. In Calibre I can group them and they show as one book. After this in calibre-web I see 2 of the book, only the first one has an image and a download link. The other is a ghost which has no image and no content. If I edit the ghost and add an image I then see 3 books - One with a download link and image, one with an image but no download link and the 3rd is a ghost again. I must be doing something wrong. This bug solution sounds like the feature in calibre but this is calibre-web issues, no?

PR #272, opened 2017-08-09
Added option to upload a new format to a given book

Kyosfonica:
Implements [219](https://github.com/janeczku/calibre-web/issues/219)

From the edit metadata form:
Added option to upload a file in the form so there is the option to have different formats of the same book.
Uploadable formats are the same as for new books entries.


jkrehm:
If you rebase onto `master` you should be able to clear out the conflict easily. Then force push to update your PR.

cps/web.py
```diff
@@ -2719,6 +2719,35 @@ def edit_book(book_id):
...
+        file_ext = requested_file.filename.rsplit('.', 1)[-1].lower()
+        if file_ext not in ALLOWED_EXTENSIONS:
+            flash(_('File extension "%s" is not allowed to be uploaded to this server' % file_ext), category="error")
+            return redirect(url_for('index'))
+
+        file_name = book.path.rsplit('/', 1)[-1]
+        filepath = config.config_calibre_dir + os.sep + book.path
+        filepath = os.path.normpath(filepath)
+        saved_filename = filepath + os.sep + file_name + '.' + file_ext
+        file_size = os.path.getsize(requested_file.stream.name)
```
jkrehm: This line caused a problem for me. When I moved it to be right before line 2743 and changed it to `file_size = os.path.getsize(saved_filename)` it worked correctly.

OzzieIsaacs:
Hi,

this looks like a promising PR. But I'm not 100% happy. If you select a 2nd file there is no feedback that which file is selected:
upload_invis
I changed the visibility of the file-button, so you can see that something is seleted:
visi

Could you please add this to your code (except the ugly button). If I understood it right than is the only posibility to add additional formats during upload of the book. Qould it be easy to add this function also to the "Edit Metadata"-page to have this function always available (it's not a precondition to have this before merge, just a question)

Kyosfonica:
I'll recheck the code. On chrome the filename was visible. this button is available both when uploading a new book and when editing metadata.

OzzieIsaacs:
Okay. On Firefox the text isn't visible as the screenshots show (or don't show)

Kyosfonica:
Reviewed on Firefox. It seems like a JS error that prevents the full js code to run.
Chrome:
image
No JS errors.

Firefox:
image
JS error at line 26 in edit_book.js.
That line, if I'm not wrong, the one that writes 'dd/mm/aaaa' inside the publishing date field.
`$("#fake_pubdate")
                .val(pubDate.toLocaleDateString(language))
                .removeClass("hidden");`

Kyosfonica:
Please, recheck this PR as I rebased the code of the branch to match master and I am not sure if I done this right. I see a lot of new files that I didn't modified.

Kyosfonica:
Please, recheck this PR as I rebased the code of the branch and I am not sure if I done this right.
In case of error I created a brand new branch with the latest update of master and just my changes.

I want to learn how to deal with this kind of situations where the master branch of the main project updates and my changes in a branched fork have to catch up.

cps/web.py
Kyosfonica: Cool, moved the line to prevent the error.

PR #279, opened 2017-08-13
Add static asset cache busting

jkrehm:
Calculates MD5 hash of static assets on server start. These are appended to the file name's when url_for is used, telling browsers to download the latest version of the files, e.g. `style.css?q=81cf0ee`.

Inspired by https://github.com/ChrisTM/Flask-CacheBust but uses query strings instead of rewriting the path because fonts inside CSS files break when the path is rewritten.

### DIFF
--- a/cps/cache_buster.py
+++ b/cps/cache_buster.py
@@ -1,0 +1,57 @@
+# Inspired by https://github.com/ChrisTM/Flask-CacheBust
+# Uses query strings so CSS font files are found without having to resort to absolute URLs
+
+import hashlib
+import os
+
+
+def init_cache_busting(app):
+    """
+    Configure `app` to so that `url_for` adds a unique query string to URLs generated
+    for the `'static'` endpoint.
+
+    This allows setting long cache expiration values on static resources
+    because whenever the resource changes, so does its URL.
+    """
+
+    static_folder = app.static_folder  # the rooted path to the static file folder
+
+    hash_table = {}  # map of file hashes
+
+    app.logger.debug('Computing cache-busting values...')
+    # compute file hashes
+    for dirpath, dirnames, filenames in os.walk(static_folder):
+        for filename in filenames:
+            # compute version component
+            rooted_filename = os.path.join(dirpath, filename)
+            with open(rooted_filename, 'r') as f:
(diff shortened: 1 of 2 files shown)


jkrehm:
The Codacy results are false positives:

1. `Unused variable 'dirnames'`: Variable is required to destructure the `os.walk` results
2. `Use of insecure MD2, MD4, or MD5 hash function.`: I went with MD5 because it is fast and we don't require a secure hash, just something to let us know the file contents changed. Chances of a a collision is both (a) extremely unlikely and (b) relatively unimportant (worst case, browser doesn't download the latest file version).
3. `Unused variable 'reverse_to_cache_busted_url'`: Function is used, but not by name.

OzzieIsaacs:
I'm not sure if everything is right. After merging to all static files a "?q=" is appended, but I don't see the calculated MD5 hashes:
grafik

Do I have to do anything for enabling it? (I also tried the hda branch in your repro withou success)

jkrehm:
Blast. Somehow I missed including the addition to requirements.txt. I'll update the PR.

OzzieIsaacs:
Yes, as you can see in the screenshot the get parameter "q" is appended to the links. Are there special requirements for the flask version to use?

jkrehm:
Not that I'm aware of. I have Flask 0.12.2 installed, but just tried with 0.11 (the minimum in requirements.txt) and it's working fine. I'm using Python 2.7.10 in my development environment and Python 2.7.11 in my "production" environment, in case that makes a difference.

Do you happen to have a debugger configured so you can step through the code in `cps/cache_buster.py`? Alternatively, print statements might be good enough. Either `hash_table` isn't getting populated, or `filename` and `file_path` aren't matching up.

OzzieIsaacs:
Okay I'll check that

OzzieIsaacs:
I found the problem it's windows vs. linux path seperators. (I'll fix it on merging)
But there is another issue:
In the epub reader in folder variable the parameter is also appended. The result is a endless spinning wheel, and the page isn't loading anymore. 
grafik

The pdf reader is working, txt reader seems to work also.

Could you please fix this.

jkrehm:
Both issues ought to be fixed now.

OzzieIsaacs:
Thanks

PR #282, opened 2017-08-15
Use lazy loading to show images

jkrehm:
Loads images when they're within 200px of being in the viewport. Prevents loading images that might not be seen, and ought to improve "time to first render" performance.

lazy-load

### DIFF
--- a/cps/static/css/style.css
+++ b/cps/static/css/style.css
@@ -15,7 +15,11 @@ a{color: #45b29d}a:hover{color: #444;}
 .container-fluid img{display:block;max-width:100%;height:auto}
 .container-fluid .discover{margin-bottom:50px}
 .container-fluid .new-books{border-top:1px solid #ccc}.container-fluid .new-books h2{margin:50px 0 0 0}
-.container-fluid .book{margin-top:20px}.container-fluid .book .cover{height:225px;position:relative}.container-fluid .book .cover img{border:1px solid #fff;/*border-radius:7px;*/box-sizeing:border-box;height:100%;bottom:0;position:absolute;-webkit-box-shadow: 0 5px 8px -6px #777;-moz-box-shadow: 0 5px 8px -6px #777;box-shadow: 0 5px 8px -6px #777;}
+.container-fluid .book {margin-top:20px}
+.container-fluid .book .cover {height:225px; position:relative;}
+.container-fluid .book .cover.has-loader .loader {display: inline-block;}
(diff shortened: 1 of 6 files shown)


jkrehm:
@OzzieIsaacs @janeczku any feedback on this PR?

OzzieIsaacs:
Let me check that lazy loading doesn't yield to endless page refreshes on e-ink readers like kindle

jkrehm:
Ah. On my Kindle Paperwhite it doesn't load the new images as you scroll down. Not sure how to debug it, but I'll see if I can figure something out.

I actually made the change primarily for my Kindle because the large number of images and their large size max out the available memory and make the browser crash. My hope is that by lazy loading I can navigate around more easily without having it crap out on me.

jkrehm:
Unfortunately the Kindle browser doesn't report its y scroll position. )= I assume its just the lousiness of the browser, because work-arounds that are supposed to work back to IE 6 still don't work.

```
$(window).on("scroll", function() {
    alert("window.pageYOffset=" + window.pageYOffset + " window.scrollY=" + window.scrollY + " document.documentElement.scrollTop=" + document.documentElement.scrollTop + " document.body.scrollTop=" + document.body.scrollTop);
});
```

All of those return `0` when I scroll on my Paperwhite, so it's impossible to know if the images have scrolled into view. Super unfortunate.

I still think lazy loading the images are an important optimization for use with normal (non-Kindle) browsers. We could do browser detection (e.g. check if `window.navigator.userAgent` contains "Kindle" and if so load all images immediately) but it does eliminate the performance (and memory) gains of lazy loading them. Thoughts?

cmer:
I'd personally make the lazy loading a lot more aggressive. It's quite annoying to constantly have to wait for images loading when scrolling. Perhaps preload within 1000px or so?

jkrehm:
I'm game for that because I've also run into the hassle of images loading too slowly.

I'm not sure what to do about browsing on Kindle. I originally created the PR because loading my library on my Kindle crashed the browser (I'm guessing it's because it runs out of memory loading all of the images) but the Kindle browser is so lousy that you can't lazy load.

So, I don't know.

cmer:
Infinite scrolling on Kindle might not be a good idea. Pagination might be better since the device has limited resources.

OzzieIsaacs:
I'm personally happy with the speed of loading the images the way it is now, so I'm hesitating (?) to merge this PR.

OzzieIsaacs:
I'll close this PR now. 
Thanks anyway for the cool javascript things you implemented it brought the project a big step ahead

PR #284, opened 2017-08-16
Meta data improvements

### LINKED ISSUE #270: Enhancement: Pull cover from online metadata if none available locally
If possible it would be good if the cover from online metadata would also be pulled if there is none available in the local source.

jkrehm:
* Add buttons to show/hide results from Douban & Google
* Trigger search with "Enter" key
* Use Underscore.js template instead of build HTML strings in JavaScript
* Keep click event handler in JavaScript instead of using HTML's `onclick`
* Normalize Douban & Google results
* Update cover image & add cover URL to the form input

get-metadata

Note: Depends on #271 since the Google Books images do not have `.jpg` in their URL.

Resolves #270


simkin:
Thank you for "Update cover image & add cover URL to the form input" !!

PR #288, opened 2017-08-19
Show "More by" on author page

jkrehm:
Uses Goodread's list of author's books, filtering out the books that are already in the user's library.

Requires the Goodreads dependency and API information.

screen shot 2017-08-19 at 10 33 24


jkrehm:
I think this PR only got partially merged. An earlier version of it was merged, but missed revisions that we made after the initial push. I've resolved the conflicts.

PR #289, opened 2017-08-19
Escape special characters in file names

### LINKED ISSUE #286: Slash in Title Metadata not being escaped
If a book has a / slash in its title metadata the name will not be properly escaped while creating the directory to put the file into. 

A "TCP/IP" for example in the title metadata will lead to a directory "TCP" being created, as well as a subdirectory with whatever follows. Calibre-web will then not be able to find the file being uploaded.

jkrehm:
Follows Calibre's convention of using underscores as replacements

Fixes #286

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -266,6 +266,7 @@ def get_valid_filename(value, replace_whitespace=True):
     """
     if value[-1:] == u'.':
         value = value[:-1]+u'_'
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
There are more special characters to escape: #*+:\"/<>? 
( I changed the comment at that point to english, so its understandable to everybody and fixed the problem by changing corresponding function call)

Issue #285, opened 2017-08-16
I want to konw the bookmark is there working when I close the website?


flashliao:
When I reed books ，not finsihed read the whole books。I click the bookmark buttons
![image](https://user-images.githubusercontent.com/28710685/29348560-2b41dc14-8287-11e7-84c3-d5922df46c71.png)
But I funod when I closed the website ,the bookmark is missing when i login in same users.
How can I place bookmark when i next time read the same books?

jkrehm:
Seems to be a duplicate of #266.

Issue #266, opened 2017-08-07
Possibility to save bookmark and adjust font size in read in browser mode


simper:
Bookmark can be created in read in browser mode but once the page is closed or refreshed, the bookmark will be lost which makes this functionality useless; and the size of the font cannot be adjusted is also an annoying thing when some of the book's default font size is too small.

fuchst:
Font size can be changed by zooming in the browser. Text reflows and it works really well. Regarding the bookmarks, this is also not working on my end. Sadly I currently don't have the time to investigate, but I assume that it just doesn't communicate with and store the state on the server. So it is only local.

OzzieIsaacs:
Bookmarks are not saved (it's just not implemented)

simper:
@OzzieIsaacs Is there near plan to implement bookmark in database for separate user?

simper:
@fuchst Thanks for the tip, on computer browser it did work well, but on mobile, it is not very convenient to do such zoom in browser, it is still appreciated if the browser js reader provides the option to do several stops of font size adjustments.

jkrehm:
@OzzieIsaacs does reader.min.js come from https://github.com/futurepress/epub.js/?

If so, it should be possible to hook into its event handler to listen for the `"reader:bookmarked"` event, e.g.

```javascript
window.reader.on("reader:bookmarked", function (location) {
  console.log(location); // e.g. epubcfi(/6/4[id7]!/4[307]/4/1:0)
});
```

We could save that to the database and append it to the URL when the user returns, e.g. http://calibre.dev.local/read/57/epub#epubcfi(/6/4[id7]!/4[307]/4/1:0).

OzzieIsaacs:
I just checked the code on their website, it is the library which is used in calibre-web

OzzieIsaacs:
I think it's a good idea to proceed as you suggested (creating a new table, holding the user_id, the books_id and the "bookmarkstring" )

jkrehm:
FYI, I've begun work on this: https://github.com/jkrehm/calibre-web/tree/bookmarks

OzzieIsaacs:
**1** bookmark per book is now saved. UI is not 100% perfect due to bugs in the underlying script used.

abhisharm:
Is this supposed to work the other way around? I have many bookmarks saved in my calibre app but when viewing them through the browser they don't show up.

PR #292, opened 2017-08-23
Save ePub bookmarks to database

jkrehm:
Save ePub bookmark to database. Also use library's built-in restore feature to restore all information from localStorage.

I haven't investigated how to hook into the PDF library yet, but its bookmarks are similar to ePub's in that they're just a specially formatted string. We won't have to change the database record to accommodate them.

Ref #266 and #285

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -1354,6 +1354,26 @@ def show_book(book_id):
         return redirect(url_for("index"))
 
 
+@app.route("/ajax/bookmark/<int:book_id>/<book_format>", methods=['POST'])
+@login_required
+def bookmark(book_id, book_format):
+    bookmark_key = request.form["bookmark"]
+    ub.session.query(ub.Bookmark).filter(ub.and_(ub.Bookmark.user_id == int(current_user.id),
+                                                 ub.Bookmark.book_id == book_id,
+                                                 ub.Bookmark.format == book_format)).delete()
+    if not bookmark_key:
+        ub.session.commit()
+        return "", 204
+
+    bookmark = ub.Bookmark(user_id=current_user.id,
+                           book_id=book_id,
+                           format=book_format,
+                           bookmark_key=bookmark_key)
+    ub.session.merge(bookmark)
+    ub.session.commit()
+    return "", 201
+
+
 @app.route("/admin")
 @login_required
 def admin_forbidden():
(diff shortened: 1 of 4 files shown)


cps/web.py
```diff
@@ -1794,49 +1814,51 @@ def unread_books(page):
 @login_required_if_no_ano
 def read_book(book_id, book_format):
     book = db.session.query(db.Books).filter(db.Books.id == book_id).first()
-    if book:
```
jkrehm: Whitespace differences because I rewrote the if-statement to be fail-fast (bail out on a failing condition). This let me remove indentation for the positive side of the if-statement.

OzzieIsaacs:
Meanwhile I had time to take a look at this PR, I found a small issue:
If you create more than one bookmark, only the last bookmark is stored to the database, whereas the UI states having more booksmarks:

bookmark

Could you please take a look at this. 
Thanks

jkrehm:
I'll check that out, but I wrote it so it clears out the current bookmarks
when you add a new one, so maybe there's a bug or something. I purposely
designed it so only one bookmark is stored.

Jonathan Rehm

OzzieIsaacs:
So only UI is the problem, it shows more than one

jkrehm:
After some debugging, I determined this is bug in the https://github.com/futurepress/epub.js/ project. It doesn't update the bookmarks each time one is added or removed. When the script first loads, it takes any existing bookmarks, loops through them, and assigns each `<li>` an ID with its current array position appended to it: 

```
listitem.id = "bookmark-"+counter
...
counter++;
```

When a new one is added, instead of re-rendering all of the bookmarks, it continues with the same `counter` variable. When a bookmark is removed, it queries the bookmarks array for the index of the bookmark, then uses that to query the DOM for the element. Since the array gets updated and the DOM elements are not re-rendered, the two become unsynchronized. For example:

1. Add bookmark A; it's given ID "bookmark-0", array index 0
2. Add bookmark B; it's given ID "bookmark-1", array index 1
3. Add bookmark C; it's given ID "bookmark-2", array index 2
3. Remove bookmark A; it's queried as array index 0, ID "bookmark-0" is removed, array is updated to remove index 0 (all indexes are updated)
4. Remove bookmark C; it's queried as array index 1, ID "bookmark-1" is removed, array is updated to remove index 1

In that scenario, bookmark B was removed from the UI instead of bookmark C. The underlying bookmarks array is correct, but the UI is out of sync.

I'm going to report this issue to the library's project. I don't know how they're going to want to deal with it, but it's not something we can fix on our end without running a custom version of the epub.js project.

PR #303, opened 2017-08-29
Fix filter regression

### LINKED ISSUE #302: Series displays random books
Since the last update to "2017-08-27T09:36:02+02:00" the Series filter doesn't work anymore, any serie of books will just display the same "random" list of books nothing to do with that serie.

jkrehm:
Re-apply `db_filter`that was accidentally removed when the language and content filters were moved to `common_filters`.

Fixes #302


jkrehm:
Thanks, @OzzieIsaacs

PR #308, opened 2017-09-02
Add progress bar to upload form

nocylah:
Add a progress bar to the upload form. The bar takes the place of the upload button when the form is submitted.

### DIFF
--- a/cps/static/js/main.js
+++ b/cps/static/js/main.js
@@ -164,6 +164,31 @@ $(function() {
             $(this).find(".modal-body").html("...");
         });
 
+    $("#form-upload").ajaxForm({
+        beforeSend: function() {
+            // Replace upload button with progress bar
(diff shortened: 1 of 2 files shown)


jkrehm:
I'm concerned about trying to replace the whole HTML document's content with the response. It's been producing JS errors for me as the `<head>` and `<body>` elements seem to get clobbered out of existence.

screen shot 2017-09-02 at 15 10 44

The response in web.py might need to be adjusted if this is going to work, e.g. be altered to expect an ajax request and respond accordingly.

OzzieIsaacs:
Good idea to give the user a feedback during the upload. 
But I have to agree to @jkrehm, it's not working properly. I also get strange errors:
error
(Firefox on Windows).
Could you please work out another solution?

Thanks

Issue #361, opened 2017-10-10
guest accounts still can not read books,this time  is no book details,I guess  program bug.


flashliao:
I doubt #360 fix the account bug,but  not notice the guest account bugs.Now book details show blank
![default](https://user-images.githubusercontent.com/28710685/31384135-319b694c-adf0-11e7-9505-631e33e3dab2.png)

flashliao:
[2017-10-10 19:32:56,934] {/root/calibre-web-master/cps/web.py:228} INFO - Starting Calibre Web...
[2017-10-10 19:32:57,655] {/root/calibre-web-master/cps/book_formats.py:18} WARNING - cannot import Image, generating pdf covers for pdf uploads will not work: MagickWand shared library not found.
You probably had not installed ImageMagick library.
Try to install:
  http://docs.wand-py.org/en/latest/guide/install.html
[2017-10-10 19:32:57,923] {/root/calibre-web-master/cps.py:28} INFO - Attempting to start gevent
[2017-10-10 19:34:26,101] {/usr/lib64/python2.7/site-packages/flask/app.py:1560} ERROR - Exception on /book/132 [GET]
Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib64/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib64/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib64/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib64/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/root/calibre-web-master/cps/web.py", line 1320, in show_book
    title=entries.title, books_shelfs=book_in_shelfs, have_read=have_read)
  File "/root/calibre-web-master/cps/web.py", line 613, in render_title_template
    return render_template(instance=config.config_calibre_web_title, *args, **kwargs)
  File "/usr/lib64/python2.7/site-packages/flask/templating.py", line 134, in render_template
    context, ctx.app)
  File "/usr/lib64/python2.7/site-packages/flask/templating.py", line 116, in _render
[...]

PR #364, opened 2017-10-11
Anonymous Browser Fix

csitko:
is_active, is_authenticated, and is_anonymous should be properties in both the UserBase and Anonymous classes to be consistent with flask_login.  Not making them properties breaks the login_required decorator in flask_login.  Additionally they're only ever called on instances of the classes so should not be classmethods in UserBase.

Previous implementation of login_required_if_no_ano decorator required the server to be restarted in order for the anonymous browsing option to take effect.

Should resolve issues #365 and #366.


gytisgreitai:
Should also solve https://github.com/janeczku/calibre-web/issues/361

flashliao:
No one commit the code?

OzzieIsaacs:
Thanks a lot, my quickfix for guest accounts was a bit buggy. I need to come up with my tests so this can't happen again.

PR #383, opened 2017-11-05
Updated Dutch translation

Thijxx:
Added missing translations and fixed some strings. Not sure if I need to generate a new `.mo` file.


OzzieIsaacs:
Thanks. 
There is no need to generate a .mo file, I do this during my build.

PR #406, opened 2017-12-10
New progress bar & IE11 fix

ytilis:
-A slimmer and more subtle progress bar to the top which only shows the page number on hover 
-Thumbnail list now scrolls to the current page on open or page change
-Added @andy29485's request for controlling page with space/shift+space
-Added the book title to the top of the page 
-Fixed IE11 failing on imagecallback

### DIFF
--- a/cps/static/css/kthoom.css
+++ b/cps/static/css/kthoom.css
@@ -1,15 +1,19 @@
 body {
-	background: #444;
+  background: #444;
   overflow-x: hidden;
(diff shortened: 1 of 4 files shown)


OzzieIsaacs:
Thanks for the work on it. 
The IE in my last commit was totally unusuable, and I also had some trouble with firefox, with out of memory errors and totally black screens. 
Before merging everything in the master branch there is some more work to do on it.

PR #458, opened 2018-03-01
Comic server

Tailchakra:
I've made comics a bit more usable for comics.

Changed Author into Publisher
Changed Series into Comics
Changed Books into Issues
Discover now shows Comics (Series) instead of Issues (Books)
When clicking on a Publisher (Author) it now shows Comics (Series) instead of Issues (Books)
Added issue number from metadata

This does require a couple mandatory metadata rules to be used optimally. But that's all possible through ComicTagger and Calibre (Best way to metadata comics anyway).

The way Calibre should be:
image

Images of the end result:

Homepage:
image

Publishers list: (Comics list is the same, but with comics instead of publishers)
image

When clicked on publisher view (DC Comics in this case):
image

Discover:
image

Issues sorted by A-Z:
image

When clicked on issue
image

It does so rating, so does the comics view. I just don't have any ratings set up myself.

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -272,7 +272,7 @@ class Books(Base):
         self.has_cover = has_cover
 
     def __repr__(self):
-        return u"<Books('{0},{1}{2}{3}{4}{5}{6}{7}{8}')>".format(self.title, self.sort, self.author_sort,
+        return u"<Issues('{0},{1}{2}{3}{4}{5}{6}{7}{8}')>".format(self.title, self.sort, self.author_sort,
                                                                  self.timestamp, self.pubdate, self.series_index,
                                                                  self.last_modified, self.path, self.has_cover)
 
--- a/cps/templates/admin.html
+++ b/cps/templates/admin.html
@@ -60,7 +60,7 @@
         <th>{{_('Calibre DB dir')}}</th>
         <th>{{_('Log Level')}}</th>
         <th>{{_('Port')}}</th>
-        <th>{{_('Books per page')}}</th>
+        <th>{{_('Issues per page')}}</th>
         <th>{{_('Uploading')}}</th>
         <th>{{_('Public registration')}}</th>
         <th>{{_('Anonymous browsing')}}</th>
--- a/cps/templates/book_edit.html
+++ b/cps/templates/book_edit.html
@@ -36,11 +36,11 @@
     </div>
 
     <div class="form-group">
-      <label for="series">{{_('Series')}}</label>
+      <label for="series">{{_('Comics')}}</label>
       <input type="text" class="form-control typeahead" name="series" id="series" value="{% if book.series %}{{book.series[0].name}}{% endif %}">
     </div>
     <div class="form-group">
-      <label for="series_index">{{_('Series id')}}</label>
+      <label for="series_index">{{_('Comics id')}}</label>
       <input type="text" class="form-control" name="series_index" id="series_index" value="{{book.series_index}}">
     </div>
     <div class="form-group">
--- a/cps/templates/config_edit.html
+++ b/cps/templates/config_edit.html
@@ -57,7 +57,7 @@
       <input type="text" class="form-control" name="config_calibre_web_title" id="config_calibre_web_title" value="{% if content.config_calibre_web_title != None %}{{ content.config_calibre_web_title }}{% endif %}" autocomplete="off" required>
     </div>    
     <div class="form-group">
-      <label for="config_books_per_page">{{_('Books per page')}}</label>
+      <label for="config_books_per_page">{{_('Issues per page')}}</label>
       <input type="number" min="1" max="200" class="form-control" name="config_books_per_page" id="config_books_per_page" value="{% if content.config_books_per_page != None %}{{ content.config_books_per_page }}{% endif %}" autocomplete="off">
     </div>
     <div class="form-group">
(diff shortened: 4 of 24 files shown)


OzzieIsaacs:
Please check:
http://jinja.pocoo.org/docs/2.10/templates/ 
Table in List of Control Structures block.
I think
{% if entry.first %}
is what you are searching...

Tailchakra:
Ha, yeah dont accept PR till I'm done with this. Good way for me to start learning the basics haha! Thought I was done but I completely forgot I still had to fix that. Sadly I'm now remotely so I'll have to push every time I want to check something.

Tailchakra:
Oh crap, should be entry.series_index.first I think.

Thing is, I cant use db.Series because series doesnt carry as much info as books do. I want to have it show the cover from the first book but sort by series. It works if I do "if == 1.0", but that wont make it show if there is no 1.0 (for example you only have 4.0, 5.0, 6.0. Rest still needs to be downloaded or whatever) So it just needs to show the first book in the index.

OzzieIsaacs:
Sorry, was wrong hint, should be:
{% if loop.first %}
Sorry

Tailchakra:
Mmh this will probably take a tad longer. I completely looked over something.

It loops all books, so it grabs a book and takes first series_index right. That's about right. But then comes the next book, but out of the same series and it grabs the first series_index out of that aswell. So it grabs the same series twice making it show up like this:

screenshot 2018-03-02 at 20 51 39

I got to figure out how to compare series so that if 2000 AD is already there, it skips. Or something similar so it only shows one 2000 AD. This hacky tacky thing might just not work out seeing as "Series" only throw id and name and not books..

Tailchakra:
Which jinja version is this? The new ninja has the unique filter which might solve this

andy29485:
So a few notes:
- The changes I made to `cps/templates/detail.html` in commit 28dac341199 seem to have gotten lost during a merge (first thing I noticed was that I had to click on 2 button instead of one to read a comic).
- From what I understand, once the comic branch matures it'll be integrated with master - so assuming every book is a issue and every series is a comic series.
  - Solution 1: Add an option to the admin panel to enable/disable the assumption that this is a comic library.
  - Solution 2: Treat only books that have the `comic` tag as issues (and only those series with at least one issue as comic series).
- Request: One of the biggest downsides of calibre-web v. calibre for comics is how hard it is to go on to the next issue/book. With novels this is not an issue (since they're long and change is less often - days vs minutes). 
  - Suggestion: once at the end of a book, a page flip will cause kthoom.js to go on to the next issue/book in the series (next nearest `series_index` - e.g. `1.0` -> `2.0`-> `2.5` -> `3.0`).

Tailchakra:
"From what I understand, once the comic branch matures it'll be integrated with master"

This might be an issue though. You dont sort comics by authors, you sort it by publisher. comicbooklover tags fit with calibre in that sense. This is why I sort by publisher and not by author (just need the latest jinja vesion for the .unique filter). Also, the way the issues are presented is completely different:

- books (title + author)
- issues (title + series #series_index)

This also is very important for comics compared to books. You cant just see books the same as comics, its an entire different scheme.

Another big thing I have against a branchmerge, is that discover will show books, discover showing issues makes no sense. You want discover to show comics instead. https://hiddenprocess.com/comics/
Only things broken are discover and publisher (when chosen a publisher) because I dont have the .unique filter. (also I broke search, fixed it locally though. Changed a bit too much)

The detail.html change was a blind change from me without looking at previous commits. This was so it works for the caliblur theme.

andy29485:
> "From what I understand, once the comic branch matures it'll be integrated with master"
>
> This might be an issue though.

Well, imagine a normal user or someone using docker. Switching branches is going to be hard for them, why not just make it a check box on the admin page? Or better yet integrate it fully with the regular book library.
- Bad Solution: Essentially, ignore everything further on, but rather than having users deal with git branches - turn this branch into a checkbox.
- For a more practical solution read on.

> You don't sort comics by authors, you sort it by publisher.

Kinda true, but this can easily be added by having another button instead of changing an already existing button. In some cases I still want to sort by author (manga). Adding another button won't be a big change and easy to integrate.
- Solution: 2 buttons (Author *and* Publisher).

> [T]he way the issues are presented is completely different: ... This also is very important for comics compared to books. You cant just see books the same as comics, its an entire different scheme.
>
> Another big thing I have against a branchmerge, is that discover will show books, discover showing issues makes no sense.

These two issues can be resolved by a similar 2-part fix.
1. First, this could be quickly solved by having every comic issue/book have the tag `comic_issue` or the such. Calibre-web could simply look for that tag and process that book accordingly.
  - e.g. If a comic is chosen for discover, pick the first issue (as you are doing now), but if it does not have the `comic_issue` tag, treat the book as before (as a book).
  - Solution: user tags comics as such OR require another calibre column that has a boolean value (true for comics, false otherwise)
2. Also, series (books) and series (comics) aren't that different. You bring the example of discovering a random issue being bad, but does that not also holds for books? (As in: do you want to start Harry Potter, Game of Thrones, or the Dune Saga from the 4th book in the series?)
  - Solution: always discover only the first book of the series (`series_index == 1.0`) whether it be a book or a comic issue (essentially we discover series not books)
    - make this the default behavior, but disable-able via checkbox (in user config or something)
    - For an upgraded version, simply pick the unread book with the lowest `series_index` in the series for discover.

> The detail.html change was a blind change from me without looking at previous commits. This was so it works for the caliblur theme.

Ah, I see. I think I should take more of an issue with caliblur then - because this is a violation of design principles, clicks should be minimized (can't be helped when there are multiple formats, but when there's only one? really?).

Tailchakra:
Im on my phone at the moment so can't give a real comment. Ive been using this docker to set up branch and even repo https://hub.docker.com/r/technosoft2000/calibre-web/ so that's why I didn't think any further.

My knowledge is not amazing so most of those things are out of my reach. I just had to change a couple things because calibre-web was virtually unusable as is for comics. Ive got nearly 250k issues organized atm (Not in calibre yet. Everything in now is just a test) and I really, really don't want to change the metadata because it's tagged as both comicrack and comicbooklover scheme.

Maybe ill have to ditch the pr then and use my own repo instead to be fully caliblur compatible and fully focus on comics (will probably have to rewrite more things and rip out book variables). Will have to kick up someone who knows more about it haha!

PR #500, opened 2018-04-22
Tags in detail view are sorted

andy29485:
A minor annoyance with a one line fix. Tags now appear in an alphabetically sorted order in the book detail view.


OzzieIsaacs:
Fixed it directly in the database query

PR #511, opened 2018-05-21
Simple Theme

otapi:
The Standard Theme without cover images and added download link buttons in listviews - handful for e-ink devices or on slow connection. Book Details (with cover image) can be opened via link of the titles.


davydov-vyacheslav:
couple cents from me:
- is it possible to provide screenshots of behavior to make more-less clear insight of how it affects UI
- maybe I am not correct, but looks like you have added code for downloading which pretty equals for _6_ pages ... , haven't you? Is there a way to extract it?
- I know that I'm nobody here .. but dozen injection of theme-specific code within many pages IMHO a bad idea. I understand that solution required architecture changes, but maybe it is possible to extract custom code and, probably, make additional folder per theme?

otapi:
Thanx for your feedback!
Well, omitting the images could be done via a simple css change, but adding the download links to the list views needs edit on all the templates where the rows are defined. Unfortunatelly, the rows are defined redundantly in 6 files. Yes, it could make simpler by defining the rows in a separate file and link it only in all the touched 6 files - but that is a bigger overhaul of the layout engine, which I did not intended for as part of this code.
Though, I made more commits into my repo in the meantime and I didn't realized gitHub adds these to this PR... So I'm closing this pull request now.

Issue #485, opened 2018-04-05
[Feature Request] Specific book deleting. 


mccorkled:
I recently noticed that several of my books have multiple copies. When I go into edit meta data and click delete book, it wants to delete all of them. Would there be a way to allow it to only delete the specific format I choose?

For instance, I have an epub and a mobi, and I just want to delete the mobi. Or I have an epub and a txt, and I just want to delete the txt. 

Thank you! :D

OzzieIsaacs:
Function is merged

PR #519, opened 2018-06-03
 Option to delete individual book formats 

inflrscns:
Displays an option on the book edit page to delete individual formats if the book has more than one. 
screenshot

When a book only has one format (or once you've deleted all the others) the page displays how it originally did:
screenshot

Implementation of this feature request: #485

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -307,12 +307,12 @@ def get_sorted_author(value):
     return value2
 
 
-def delete_book(book, calibrepath, format=None):
+def delete_book(book, calibrepath, book_format=None):
     if "/" in book.path:
         path = os.path.join(calibrepath, book.path)
-        if format:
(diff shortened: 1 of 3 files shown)


davydov-vyacheslav:
Just wondering, is there a way / how assign several book formats to book via UI?
and one more question ... maybe it is offtopic, but is there a way to assign several languages to the book (e.g if we have book in native language and its translation, and its quite good to manage them under one 'book' entity)?

OzzieIsaacs:
You can assign several book format to one book by uploading them in the edit books section, otherwise currently not. (#219 requests this feature)
Regarding several languages request, I think Calibre can‘t handle this, you can have several book formats in one book, but you can‘t determine that this book format has this language and the other has a different language. You need to have 2 books.

Issue #523, opened 2018-06-14
Nginx Proxy issue


Fmstrat:
Two items:

1) The nginx configuration in the README has some errors when used as-is. Right now, it's:

```
http {
    upstream calibre {
        server  127.0.0.1:8083;
    }
    server {
            location /calibre {
                proxy_bind              $server_addr;
                proxy_pass              http://127.0.0.1:8083;
                proxy_set_header        Host            $http_host;
                proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_set_header        X-Scheme        $scheme;
                proxy_set_header        X-Script-Name   /calibre;
        }
    }
}
```
But it should be:
```
http {
    upstream calibre {
        server  127.0.0.1:8083;
    }
    server {
            location /calibre {
                proxy_bind              $server_addr;
                proxy_pass              http://calibre;
                proxy_set_header        Host            $http_host;
                proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_set_header        X-Scheme        $scheme;
                proxy_set_header        X-Script-Name   /calibre;
        }
    }
}
```
Notice the `proxy_pass` line and it using the `upstream` name of `calibre`.

But, even with this change, there's a problem. If this is running on a non-standard port, Calibre-web is responding with headers that are causing nginx to redirect out. For instance:

```
        server {
                listen 12345;
                server_name HOST.DOMAIN.com;
                gzip off;
                ssl on;
                ssl_certificate /letsencrypt/live/DOMAIN.com/fullchain.pem;
                ssl_certificate_key /letsencrypt/live/DOMAIN.com/privkey.pem;
                ssl_session_cache shared:SSL:10m;
                ssl_protocols TLSv1.2;
[...]

PR #528, opened 2018-06-15
Readme update for nginx proxy

Fmstrat:
https://github.com/janeczku/calibre-web/issues/523


OzzieIsaacs:
I could not reproduce everything, and added only the ssl hint and the max filesize

PR #530, opened 2018-06-16
Make cover image bigger in details view

otapi:
Make cover image bigger in details view at mobile devices


OzzieIsaacs:
Thanks for the PR. Meanwhile I had time to check it. Here are the detail views at my phone and my reader before:
before_phone
reader_before
And with your PR:
after_phone
reader_after

I don't see a advantage with the bigger cover-pictures. I think it forces you to scroll more to see what you might want to do, so I decided not to merge it. Sorry

PR #618, opened 2018-09-10
Enhancing the update check

vigri:
Hello,

this PR updates and extends the update check in calibre-web.

- check for newer versions is now more robust regarding possible errors
- error / success / info messages
- displaying "what's new" in the more recent versions than the current one
- fix html/css errors

no-update

update-available

update-failed


vigri:
The typo in the first image "Now update...." is fixed in the last commit ;)

OzzieIsaacs:
Thanks a lot for the contribution. I merged it, but I reverted the change with the version file 
(I don't like having dozens of files on my harddisk)

bodybybuddha:
Really nice contribution!

PR #629, opened 2018-09-15
Audiobook support -- updated with code review

Mohammed90:
This build on top of #495 incorporating the review comments from the maintainers.

@OzzieIsaacs could you please test it locally or on some clean environment? I'm having issues with running it locally (pip, blah...). I'm not a Pythonista, so debugging pip issues is a bit hard for me.

PR #495, opened 2018-04-16
Added audiobook support

modembug:
Adds support for audiobooks. Includes functionality to sync progress of built in player to bookmark entry. This allows users to listen on multiple platforms and maintain position.

image

image

PR #634, opened 2018-09-16
Audiobook Feature

bodybybuddha:
The PR is an updated fork of @Modembug's initial audiobook feature [PR #495](https://github.com/janeczku/calibre-web/pull/495).  

It contains the suggested fixes from that PR, which @mohammed90 had cleaned up in this [PR #629](https://github.com/janeczku/calibre-web/pull/629).

I'll keep this branch up to date so that we don't fall into the same situation we did with the original PR.

I've completed some testing - primarily on a Mac with Safari, Chrome, Firefox.  Testing on other devices and platforms are requested!

Note - @Modembug:  I had to change the details page icons code "slightly".  I wasn't able to cleanly see all the options.  The current implementation may not be elegant - but it appears to work.  I think that maybe we allow for additional playback types in the future.  Maybe drop soundmanager2 and go with another library?


modembug:
Sounds good to me. This implementation was only done up to suit my particular requirements and I am very new to Python so I probably made a mess. :)
________________________________

Mohammed90:
The image files should be moved to `images/` instead of `image/`

bodybybuddha:
@Mohammed90 - yeah, I sorta skipped that one.  Thanks for catching it.

I've made the change just now.  However, I don't really think this was a necessary change as we already had images in multiple locations.

Mohammed90:
Honestly, I'm not familiar with the project's file organization. This was just one of the review comments on the first PR.

bjhiggins:
The last time I played with this code, I found that it *would* play .M4B files.  There may be other extensions it will play as well.

Additionally, if the book cover image was huge, the player controls were unavailable.  I made those changes eons ago - but could never get them into the PR and abandoned and deleted them :(

bodybybuddha:
@bjhiggins - Ah okay - no worries.  Thanks for replying.

Soundmanager2 does support more than just mp3's.  It looks to play a bunch of formats.  I was debating if it was worth keeping it or just use just HTML5 (why not if we are only going to serve up mp3s?)  But looking at the source files, it appears to handle a few more formats that HTML5 doesn't natively support.

I'm working on some language stuff for the site now, but I'll take a look at expanding the format types afterwards if no one else takes it up.

modembug:
With the Sound manager 2 library I was able to hook into the position of the stream and write it back to the bookmarking interface in calibre-web. I think if HTML5 can do that, it could be replaced.
________________________________

bodybybuddha:
Added the ability to play any of the uploaded audio book formats.

Looking into @modembug's comment on hook the position of the stream.  I must have missed that functionality.  I'll look into that next.

bodybybuddha:
Okay - looks like the calibre.web javascript block needed some fixing.  It is now rendering correctly and I've tested the bookmarking - works well @modembug!

I also removed tax format from the app.  We can add it back if there is a big demand for it.  I don't have any files to test with, so I just removed it.

Mohammed90:
Given your experience with the codebase, how complex do you think it'll be to allow for an audiobook to consist of multiple files?

bodybybuddha:
TL;DR - IMHO, Not easy.  But something I would love to see done.

The answer itself is complex as there are a number of factors to think through and make decisions on.  Technically its easy as we're the ones writing the code - however this comes at a cost of flexibility of configuration and severely limits the use cases for CW.

Here is a [link](https://www.mobileread.com/forums/showpost.php?p=2141857&postcount=28) that describes the known ways of doing this within calibre.  I'm basing my answer off these methods.

First, the philosophy of calibre is one **ebook** = one file.  As CW is using the calibre db as a base, a lot of what we do is influence by this philosophy.  All of the solutions that people are using for adding multiple audio book files, or even adding an a single file audiobook, to calibre are exploiting the original design/purpose.  Even multi-user access to the system isn't fully baked in (and probably never will be.)

So to do multi-file audiobooks, we would have to decide a number of things.  (just some of the things - I'm sure there are more): 

- Location of files - Do we store the audio files in the library - or do we just have a pointer in the library to where the files live?  This will affect all areas of uploading and basic I/O for these files part of the library or stored outside of the library.
- Metadata - We need to decide what metadata to keep for these files.  Then do we read the metadata on upload?  This could be quite a bit of code.  The link above is to a thread of the author of a plugin for calibre that does this sort of work.  We aren't reading any metadata now for audiobook uploads.
- User Access -  Fortunately, Modembug's selection of SoundManager2 is good one, as it allows for creating a play list - thus listening on a browser is taken care of.  How would we offer the option of downloading?  

Please don't think I'm being negative - I would really like to see the feature.  There are just a lot of decisions to be made.

Honestly, I was thinking that we just bite the bullet and do it.  My users only use CW so my focus is to make their experience seamless.  Download, listen from a browser are done - but I'd like to see  @felixscheinost's idea of an RSS feed (for podcast players, etc.) I'm the only one that uses calibre to manipulate data, etc. so I wouldn't care that I couldn't use all of the features of calibre.

I would propose we write an entry in calibre db for any audiobook that will trigger CW to look into its own database for audiobook information like the location of files.  Metadata would be split between calibre db and CW db - for instance "Read by" would be in the CW db.  Then we can just address the GUI items.  I think this gives us the ability to add the feature without mucking around with all the other exploits people have devised (series hacks, custom columns, different databases, etc.) This will cut down on the number of dependancies we would need to enforce in order to support those exploits.  Of course, if we go this direction, you really can't use calibre application to access the audiobook files.

Mohammed90:
Thank you, @bodybybuddha , for the detailed response!

bodybybuddha:
Was trying to figure out how to get audiobooks to be downloaded to IOS devices.  Turns out I had to change the MIME type to application/audioext as opposed to audio/audioext.  Now, if you use an IOS app that has a builtin browser (for example, the extremely flexible Documents IOS app), they will be able to properly download the file.  

Documents will also play the audiobook - or you can move it to whatever service you want, iCloud, Google drive, Dropbox, whatever Documents support.  Then, use another audiobook app to listen to the audiobook.  For instance, the Bound IOS app is a very good player and can communicate with those cloud services and download the audiobook in it's ios 'sandbox'.

benklop:
@bodybybuddha looks like there are some merge conflicts here. most of them look pretty trivial, but I'd like to test this feature out and provide feedback.

OzzieIsaacs:
I‘ve already merged it to the development branch, on need to change anything here. It‘s Wasted time

benklop:
ok, cool. i didn't realize there was already a branch with the conflicts resolved. awesome.

OzzieIsaacs:
The code is merged now, but I merged it into the develop tree and now merged it back to master, github didn't realize this. So i close it.

Dinth:
Hi. Sorry for refreshing the old thread but how exactly (multifile) audiobooks should be imported to calibre? Im not seeing anything in the documentation and when im trying to import a ZIP file containing mp3 files im getting an error saying that the server doesnt accept this format.

Issue #621, opened 2018-09-10
Login screen is always english


OzzieIsaacs:
The login screen should get the locale settings from the browser, currently this is not working, and english is always shown.

OzzieIsaacs:
Merged

PR #635, opened 2018-09-17
Fix for #621 - Login page always in english

bodybybuddha:
The Guest user is configured for 'en' language by default.  The get_locale function need to include a check for this user.

Now the login page page will now render the in the browser's preferred language.  In addition, anonymous browsing will also be in the browser's preferred language.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 helper.py ident export-subst
+/test export-ignore
 cps/static/css/libs/* linguist-vendored
 cps/static/js/libs/* linguist-vendored
--- a/cps/book_formats.py
+++ b/cps/book_formats.py
@@ -123,9 +123,11 @@ def pdf_preview(tmp_file_path, tmp_dir):
 
 def get_versions():
     if not use_generic_pdf_cover:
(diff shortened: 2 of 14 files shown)


OzzieIsaacs:
Thanks for the fix. While checking it I realized one thing: IF anonymous browsing is activated the admin can set a language for the guest user, which is now no longer taken into account. I want to think a bit on it, but currently I tend to removing the language setting for the guest user, so having a consitent behavior across the guest user usecases

bodybybuddha:
Thanks for taking a look.  

In my mind, if anonymous is activated, I would assume the administrator really doesn't care about the language the user is rendering the site in.  I think it that by not forcing a language, it gives the site more flexibility.

However, if you think it's important, we could just add a "force language" option to the config?

OzzieIsaacs:
I‘m currently writing automatic test routines I currently feel how much work and complexity every option creates. So I decided to keep it as in your pull request.  (No need for you to merge it to the newest master commit, l‘ll do that tomorrow)

bodybybuddha:
Cool!  I was just going through all my branches and making sure they were all up to date with the master branch, hence the update.

PR #651, opened 2018-10-01
Beautify http errors

vigri:
Hello,

this PR beutifies the http errors.
Instead of the standard "not found" etc. page we get a nicer one.

@OzzieIsaacs should be localize the error messages too?
Right now the english message comes directly from werkzeug.HTTPException

cw_404
cw_503

### DIFF
--- a/cps/static/css/style.css
+++ b/cps/static/css/style.css
@@ -5,6 +5,22 @@
   src: local('Grand Hotel'), local('GrandHotel-Regular'), url("fonts/GrandHotel-Regular.ttf") format('truetype');
 }
 
+html.http-error {
+  margin: 0;
+  height: 100%;
+}
+.http-error body {
+  margin: 0;
+  height: 100%;
+  display: table;
+  width: 100%;
+}
+.http-error body > div {
+  display: table-cell;
+  vertical-align: middle;
+  text-align: center;
+}
+
 body{background:#f2f2f2}body h2{font-weight:normal;color:#444}
 body { margin-bottom: 40px;}
 a{color: #45b29d}a:hover{color: #444;}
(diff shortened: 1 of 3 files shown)


OzzieIsaacs:
I'm a b it afraid that adding additional dependencys to the error page this could be a new source for errors, so I'd prefer not to translate them.

vigri:
Good point. I think we should let the messages in English too.
In an ideal world the user shouldn't see an error message anyway 😆

OzzieIsaacs:
Question: In case of an error 500 (the most occouring page), do we also see this page in future, or is this page unchanged? (for the error 500 the message should be different from now, because it's never an overloaded server, it's always an application error)

vigri:
This page should only be shown when an `abort()` has been called.
Does this answer you question or did you mean something else?

bodybybuddha:
@vigri - Sorry, I'm relatively new to python & flask, but I had a question about this implementation.

I understand that your implementation would register a generic http error function to all exceptions in the underlying werkzeug library.  

However, doesn't flask already have this builtin with @app.errorhandler?  For instance we could have two - one for routes and one for everything else:

```
@app.errorhandler(500):
    return render_template('http_500_error.html',
                            error_code=error.code,
                            error_name=error.name,
                            instance=config.config_calibre_web_title
                            ), error.code

@app.errorhandler(Exception):
   return render_template('http_error.html',
                            error_code=error.code,
                            error_name=error.name,
                            instance=config.config_calibre_web_title
                            ), error.code
```

Just trying to understand if I'm missing something.  Thanks!

OzzieIsaacs:
Thanks a lot.
I changed it a little bit, as the error 500 was included in the error-handling. For a error 500 a new exception was thrown. 
Because in case of an application error the exceptions had different attributes.
To answer @bodybybuddha 's question: The error function is called for all error 400 class errors which are more than 20 (https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors), handling it this way is much shorter than having a function with over 20 decorators

Issue #295, opened 2017-08-26
send azw3 to kindle email


jiangjiangflies:
Hi,
Why I can't send azw3 format ebook to  my kindle email?

jkrehm:
Haven't tried it myself, but Googl'ing for "send azw3 to Kindle email" makes me think that it's an issue with the DRM that's part of AZW3 files.

jiangjiangflies:
Can't kindlegen convert azw3 to mobi or epub before sending emails?

jkrehm:
I don't have a file handy. Have you tried it with your copy of kindlegen? I don't believe it will because azw3/kf8 is designed to prevent viewing on non-Kindle devices. 

Personally I use a slightly older version of Kindle for Mac (1.17, I believe) so that I don't have to deal with KF8.

jkrehm:
Ah nevermind, KFX/AZW6 is the problem, not KF8/AZW3. So at least the DRM shouldn't be an issue. Sorry for wandering into the weeds.

jiangjiangflies:
I think it would be better if you can replace kindlegen in the vender with ebook-convert binary.

jkrehm:
Can you provide some details on what error you get? Have you tried another ebook? Have you tried sending the file directly to your Kindle email (not via the calibre-web interface)?

jiangjiangflies:
I tried to send the file directly to my kindle email, amazon refused......It  doesn't support azw3 file to mail. So  could you add a function to convert azw3 to mobi before sending?

jkrehm:
`ebook-convert` does seem like a good way to go to support a wider variety of book formats. It requires installing Calibre, which may or may not make sense, e.g. I run calibre-web on a headless (no GUI) Linux server.

So if `ebook-convert` is made to be an option, I think `kindlegen` needs to remain.

For now, I recommend you use Calibre to convert your AZW3 books to ePub or MOBI format. calibre-web's code doesn't even support AZW3 files, so I don't know how it even mailed the file out.

```
    for entry in data:
        if entry.format == "MOBI":
            formats["mobi"] = os.path.join(calibrepath, book.path, entry.name + ".mobi")
        if entry.format == "EPUB":
            formats["epub"] = os.path.join(calibrepath, book.path, entry.name + ".epub")
        if entry.format == "PDF":
            formats["pdf"] = os.path.join(calibrepath, book.path, entry.name + ".pdf")

    if len(formats) == 0:
[...]

PR #655, opened 2018-10-01
295   send azw to kindle

bodybybuddha:
Here is my interpretation of a fix for #295.

Added a pre-render check to ensure if we should expose the Send to Kindle button.  The check is based on what converter is configured (if any) and what available formats are currently in the system for that book.

Small bug fix to ensure that any new book added to the calibre database has a format field that is all capital letters.  

If a book now has any of the following formats, the system will send a mobi to the kindle:  ePub, azw, azw3.

I don't have any azw files - but I did test this with a book that only has azw3 version and it works.  The assumption I'm making here is that ebook-convert can convert an azw file to mobi.  I couldn't find if it could in the ebook-convert documentation.

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -185,6 +185,10 @@ def send_mail(book_id, kindle_mail, calibrepath, user_id):
             formats["epub"] = entry.name + ".epub"
         if entry.format == "PDF":
             formats["pdf"] = entry.name + ".pdf"
+        if entry.format == "AZW":
+            formats["azw"] = entry.name + ".azw"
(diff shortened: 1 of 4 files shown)


bodybybuddha:
.... one day (today) I will learn to configure my code editor to remove trailing whitespaces ....

OzzieIsaacs:
I changed the function, so azw3 is now the preferred format to send it to Kindle. And I also included a dropdownlist so users can choose which format they want to send. Thanks for work you put in.

Issue #617, opened 2018-09-08
Texts from background tasks are not translated


OzzieIsaacs:
The status texts finished, started, failed and so on are not translated, because they are generated in the background task. In the background task no information on the current locale is present. The texts have to be generated in the UI:
- javascript/html, or in 
- python main thread during send to the client)

bodybybuddha:
LOL - I read this the wrong way.

Got the text to render in the correct language after a very long struggle with serializing JSON data.  It was glorious!  It worked perfectly for end users.  Then I realize that if you have admin rights, the texts were in the wrong language as they were rendered in the language of the user.  

So looking into resolving this bug the right way now!  Keeping the status codes as they are and I'll used them to render the correct verbiage in the main thread.  I'll have to think about the non-status text though.

bodybybuddha:
Figured out the issue with the status messages and multiple users with different languages!  I was only copying the task list object as opposed performing a deepcopy.  So I was changing the original object - rookie mistake!

Still need to figure out how to get the actual task text to be converted.  Trying to avoid using the javascript method.

bodybybuddha:
@OzzieIsaacs - I think I have a proposal for handling the localization of the task table. I think we need to restructure the table a bit.  As this would require a different way of using the table and would required a log o changes, I'd like to get your thoughts on the below before starting:

**Existing table:**
![image](https://user-images.githubusercontent.com/2854558/45768642-76e23b00-bc0b-11e8-9753-e0c59adac1ec.png)

Take the existing "Task" column, where we are storing text that can't be localized very well, and break it up to two columns:  Task and Message.  Where Task would be that actual task (this can then be localized very easily) and Message would be any wording we would consider useful in explaining the Task.  For example, Task = 'Convert' and Message` '<title> <sourceformat> -> <targetformat>'`.  The key is to keep the actual text of the "Message" to a point where we don't need to localize.

This would require quite a few changes to the existing code, however, I think it gives us a better, more structured way forward to handle any future request types and handle localization as much as possible.

OzzieIsaacs:
@bodybybuddha Sorry for the late reply, I deleted the notification email.
I took a look at the changes you made so far. It looks like I thought it could be solved. In my opinion, there is no need to change the table. If we have just the "raw" informations (coded in numbers or so) in the thread and if the page/infos are requested the main routine is calling the task status and creating the translated view like:
route(/tasks)
def show_view():
 variable=other_thread.get_task_status()
 ret_ele=list()
[...]

PR #668, opened 2018-10-03
Fix for #617

bodybybuddha:
For the Task table, I've localized the Status column and the Task column.

Also cleaned up the usage of 'typ' and 'type' with regards to Tasks and Worker thread.  They have been renamed to taskType and taskMessage to clear up any confusion.

I did NOT update messages.po for the new text strings that are needed.  @OzzieIsaacs could you do that?

I did test this fix using two machines - one with English-based user and another with a German-based user.  During the conversions of books and email tasks, I was able to switch between the two machines and see that the exposed information was rendered correctly.

Let me know if there is anything else we need to do here.

Thanks!


OzzieIsaacs:
Thanks a lot, I added translation functionality for some missed strings (the subject and body of the send e-mail)

PR #700, opened 2018-11-14
Update Simplified Chinese translation. Fix #681 in the case of Simplified Chinese

### LINKED ISSUE #681: Guest account  show English language and can not change it.
Guest account  show English language and can not change it. 
I want change chinaese language   test by lastest version.

admin account   or other account  can change language.

idalin:


OzzieIsaacs:
stupid question: Can‘t we just change the translation to zh_CN, if we have traditional chinese at some time we can use zh_TW?
So we can avoid this problem

idalin:
> stupid question: Can‘t we just change the translation to zh_CN, if we have traditional chinese at some time we can use zh_TW?
> So we can avoid this problem

it seems babel does not use 'zh_CN'. I had tried that  my first translating. I am not familiar with babel, but i will keep trying to improve this.

OzzieIsaacs:
I did also a little bit of internet research, looks like the behaviour is intended. 
Anyway, it's merged and fixed for now

PR #706, opened 2018-11-19
A better solution to #681

vrabe:
Babel supports likely subtags (see [here](http://babel.pocoo.org/en/latest/locale.html#likely-subtags)).
We just need to process Accept-Language values using Locale.parse().


OzzieIsaacs:
Thanks a lot. 
It helped me to understand how it's supposed to work.

Issue #701, opened 2018-11-16
Error when add/change language editing metadata.


JerryC75:
Hi! I'm having these error when trying to add or change the Language field editing metadata.
I'm using Spanish language.

![deepinscreenshot_select-area_20181115182100](https://user-images.githubusercontent.com/45083473/48590701-284cd680-e906-11e8-8552-86ac5a5b4286.png)

![deepinscreenshot_select-area_20181115182314](https://user-images.githubusercontent.com/45083473/48590706-300c7b00-e906-11e8-89c6-6e5932d71491.png)

This is the result when query the languages:
https://www.fakeweb.com/calibreweb/get_languages_json?q=
` [{"name": "Romansh"}, {"name": "Scots"}, {"name": "Sicilian"}, {"name": "Romany"}, {"name": "Romanian"}]
`

This is the result in the log file:
```
 [2018-11-15 17:29:18,048] {/var/www/calibre-web/cps/web.py:3770} ERROR - 'Languages' object has no attribute 'name'
 Traceback (most recent call last):
   File "/var/www/calibre-web/cps/web.py", line 3733, in edit_book
     modify_database_object(input_l, book.languages, db.Languages, db.session, 'languages')
   File "/var/www/calibre-web/cps/web.py", line 604, in modify_database_object
     elif db_element.name != add_element:
 AttributeError: 'Languages' object has no attribute 'name'

```

OzzieIsaacs:
I couldn't reproduce it, but I found the problematic line and fixed it.

PR #731, opened 2018-12-24
Fix adding an existing language to a book

grunjol:
When I am trying to add an existing language to a book I am getting something similar to #701 
The problem in the case of language is the element exists so 

https://github.com/janeczku/calibre-web/blob/677264950d75ec9cab9e131e800218912d5f1304/cps/web.py#L586

Its false.

Then you are checking for 
https://github.com/janeczku/calibre-web/blob/677264950d75ec9cab9e131e800218912d5f1304/cps/web.py#L597

and in the case of language is also false, so you fall into

https://github.com/janeczku/calibre-web/blob/677264950d75ec9cab9e131e800218912d5f1304/cps/web.py#L601

but the language doesn't have .name attribute and the exception rises.

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -580,24 +580,29 @@ def modify_database_object(input_elements, db_book_object, db_object, db_session
                 db_session.add(new_element)
                 db_book_object.append(new_element)
             else:
-                if db_type == 'custom' and db_element.value != add_element:
-                    new_element.value = add_element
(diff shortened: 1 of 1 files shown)


grunjol:
Bonus fix!
Only the first custom column was processed on edit, the return was set inside the loop discarding the changes in the rest of the columns

PR #765, opened 2019-01-11
caliBlur fixes

hexeth:
- Fix for theme not applying correctly when not used in reverse proxy
- Code clean-up
- Fix improperly minified CSS
- Remove "Beta" from theme title in selection


leram84:
> Remove "Beta" from theme title in selection

💯 :) :) :) almost a year and a half for that!!!! Feelsgoodman

OzzieIsaacs:
Okay, we got it, finally.
Thanks a lot for all the work you put in (and maybe all the work to come in the future). There's one point left, which is more on my side, you introduced some texts which aren't included in the original theme (like account, mark read,...) I will find some place to add the texts, get them translated and replace the tooltips (I did already for some texts I found)

PR #783, opened 2019-01-26
Bugfix

artiomn:
Code doesn't work with non-ascii filenames and paths.

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -448,20 +448,25 @@ def save_cover(url, book_path):
 
 
 def do_download_file(book, book_format, data, headers):
+    enc = 'utf-8'
+    book_format = book_format.encode(enc)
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Could you please provide a non working directory and file name?
I checked it with (running Linux, python 2.7 and python 3.6):
/bücher/从萨иownиkku/从萨и002ange19110243204 (219)/从萨и002ange19110243204 - 从萨иownиkku.pdf
and everything worked fine

artiomn:
Sorry, I can't reproduce it, but it had happened, during my experiments with Calibre: non-ascii symbol was stored in the `name` field of the `data` table.
With this patch, all will work fine regardless of the symbols type.

OzzieIsaacs:
Problem not reproducible for me

PR #823, opened 2019-02-22
Force namespace resolution of imports

AngelByDay:
The current import for some local modules causes errors when those already exist in site-packages on the target system.
Namely "import db" is ambiguous and may cause errors due to loading of the 'db' PyPI package instead of the module.


OzzieIsaacs:
Thanks, I corrected it also in some other files

PR #826, opened 2019-02-25
Shorten the display of authors if there are too many

vigri:
Hi there,

I've added a book to calibre-web which has many authors.
While this works technically perfect, it will "break" the layout.

Before PR

This PR introduces a function to calibre-web which shortens the display of authors if there are too many.
The default is set to 5 authors.

Reducing in action

By clicking on the 3 dots the whole list with authors will expand. Clicking on "reduce" the list will be reduced again.

The behavior can be modified in "UI configuration":

UI configuration

Setting the value to 0 will deactivate the "reducing" which results in the same behavior like calibre-web is acting now.

### DIFF
--- a/cps/static/css/style.css
+++ b/cps/static/css/style.css
@@ -36,6 +36,11 @@ a{color: #45b29d}a:hover{color: #444;}
 .container-fluid .book .meta .title{font-weight:bold;font-size:15px;color:#444}
 .container-fluid .book .meta .author{font-size:12px;color:#999}
 .container-fluid .book .meta .rating{margin-top:5px}.rating .glyphicon-star{color:#999}.rating .glyphicon-star.good{color:#45b29d}
+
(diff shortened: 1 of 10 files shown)


OzzieIsaacs:
Could you please do me a favor and check if the view is also okay if you are using the Caliblur! theme.

vigri:
Sure.

Well it doesn't braekes up anything in the Calibur!Theme ;)
Calibur!Theme seems to shorten the line after 30 chars.

With short names there is a benefit and this PR works like in the normal theme:
img1

img2

Using the "..." on the left book the list gets expanded.
In real world (the book on the right side) the author line is probably already to long when you have multiple authors and 3 dots will be appended.

To sum up:
Yes it works with Calibur!Theme but there isn't a real benefit.

The question is how we should deal with this. Only enable this feature for the default theme?

Issue #843, opened 2019-03-15
[Enhancement] Allow upload of png files via URL and local drive for covers


ariesdevil:
For now cover url just support jpg: https://github.com/janeczku/calibre-web/blob/a0a15b4598d372b05cb30c06ac5787dc229b7e9d/cps/helper.py#L446-L450

Oreolek:
The calibre database uses JPEG for covers. calibre-web does have a backward compatibility to calibre. https://manual.calibre-ebook.com/_modules/calibre/db/cache.html#Cache.cover

ariesdevil:
How about allowing to recevie png and convert to jpg?

OzzieIsaacs:
Your last proposal sound reasonable. I changed the title accordingly.

ariesdevil:
@OzzieIsaacs I create a PR for it.https://github.com/janeczku/calibre-web/pull/865

cbartondock:
This was closed but not implemented correct?

ariesdevil:
@cbartondock What problems are you having?

PR #865, opened 2019-03-26
Support issue: #843

ariesdevil:
Not support svg right now. https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html


OzzieIsaacs:

ariesdevil:
Thanks.

PR #868, opened 2019-03-29
Updated epub.js to v0.3.66

pwr:


OzzieIsaacs:

PR #869, opened 2019-03-29
support pycountry if iso-639 is not available

pwr:
Some Linux distributions do not provide iso-639 as a system package (Debian at least).
This patch uses pycountry as a fallback, is iso639 cannot be imported, making it easier to run calibre-web without the use of pip.


OzzieIsaacs:

PR #872, opened 2019-04-03
Use comicapi to enable basic metadata reading for CBZ/CBT files

wildthyme:
Basic metadata (title/author/cover-if-jpg/etc.) is now extracted from comics if it's in a format comictagger supports. CBR's are still disabled, mainly because libunrar is a hassle and a half.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,3 @@ gdrive_credentials
 
 vendor
 client_secrets.json
-
-# External dependencies - need proper integrating
-libunrar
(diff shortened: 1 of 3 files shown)


OzzieIsaacs:
Would it be possible to include the new dependency comicapi into this repro instead of using a half offical github repro? I would feel more comfortable doing it so.

wildthyme:
it's possible, but it would be bad for modularity and seems like a violation of best practices to me. of course you don't want it loading code that i've got the ability to change on a whim, but for that reason it's locked to the specific git commit of comicapi.

PR #908, opened 2019-05-06
Full screen fix + Reading direction + Japanese translation

subdiox:
# Changes
- Add reading direction setting to reader page
- Fix an issue that full screen button doesn't work on chrome due to its old version of script
- Fix Japanese translation file

# Description
## Reading direction
In Japan, we read comics from left to right, not from right to left. That's because raw Japanese comics (manga) is vertical writing. I think calibre-web also needs to add settings to fulfill that demand.

## Full screen issue
`screenfull.js` in the original calibre-web is too old, and it's out of date. It needs to be updated to the latest version in order to make it work on chrome or other browsers.

## Fix Japanese translation file
The original Japanese translation was too strange for me (native Japanese), so I re-wrote it entirely.


OzzieIsaacs:
Thanks a lot especially for the translation upddate

Issue #456, opened 2018-02-17
Setup.py


blade2005:
I was curious if there was a reason why a setup.py file isn't included?

Most projects I see do have a setup.py this aids in the install process and installs the script entry point in the correct ${PREFIX}/bin dir.

This would also allow the project to get onto pypi

decentral1se:
I'd be willing to pick this up if someone is interested in reviewing the PR? With the latest modern setuptools situation, the `setup.py` is not the nasty monster it used to be. It would be really great to get #677 then moving along afterwards.

OzzieIsaacs:
I'm interested in the feature, if you could implement it, this would be great

decentral1se:
OK, I've opened up https://github.com/janeczku/calibre-web/pull/927 for it :+1:

Issue #677, opened 2018-10-14
Calibre-Web Pip packaging 


arky:
Please kindly create pip packages for calibre-web. It will help wider distribution and adoption of the software.  Installing calibre-web would as simple as running:

`'pip install calibre-web' `

It would be helpful to have development version of the pip package to do pre-release testing.

`'pip install --pre calibre-web' `

decentral1se:
I'm pretty interested in this and have left a comment over at https://github.com/janeczku/calibre-web/issues/456#issuecomment-491264590. Please let me know what you think. If we had a sufficient `setuptools` integration you would not be against uploading the package to PyPi @OzzieIsaacs?

decentral1se:
Working towards this in https://github.com/janeczku/calibre-web/pull/927 :+1:

decentral1se:
Btw, we can use https://test.pypi.org/ to test we're getting everything right before making an official PyPi release. More information about that over in https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives.

decentral1se:
OK, #927 is merged. How to move forward on this? As per https://github.com/janeczku/calibre-web/pull/927#issuecomment-497734158, the release workflow is done somewhere else, so I am not sure how I can support it. I am https://pypi.org/user/decentral1se/ on PyPi if you need additional collaborators.

epsilon-0:
@OzzieIsaacs necro bumpin.  
Is this still a thing that you are trying to get done?  
I'm packaging this for OpenBSD and it would be good to get this upstream instead of maintaining downstream patches.

OzzieIsaacs:
It's still open: The problem so far is that via pypi you can not install dependencies via git, which is possible via the normal command line. Comicapi is causing the problem so far. Meanwhile I changed the code so much, that I'm thinking I'm having the right to publish it on pypi by myself, this would solve issue. But will still take some time

epsilon-0:
I've created a pull request which kind of somewhat does the packaging. I've not implemented the dependency calculation and things of that form, please feel free to add those.
#1459

OzzieIsaacs:
The project is now available on pyi under https://pypi.org/project/calibreweb/

PR #927, opened 2019-05-31
Add initial setuptools integration

decentral1se:
Closes https://github.com/janeczku/calibre-web/issues/456.

Given a recent version of setuptools (pip install -U pip setuptools), you can now do:

* `pip install -e .`
* `pip install -e .'[gdrive]'`
* `pip install -e .'[goodreads]'`

Questions:
  * Did I get the right dependencies in the extras (from optional requirements.txt)?
  * Would you also like to integrate https://github.com/pypa/setuptools_scm/ for versions?
    * Needed for https://github.com/janeczku/calibre-web/issues/677 or some workflow.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,3 @@
+include cps/static/*
+include cps/templates/*
+include cps/translations/*
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,62 @@
+[bdist_wheel]
+universal = 1
+
+[metadata]
+name = calibre-web
+url = https://github.com/janeczku/calibre-web
+project_urls =
+  Bug Tracker = https://github.com/janeczku/calibre-web/issues
+  Release Management = https://github.com/janeczku/calibre-web/releases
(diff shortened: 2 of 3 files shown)


decentral1se:
Hmmm, I'll need to investigate if all the theme/static/etc. files are being built into the dist.

OzzieIsaacs:
Regarding the dependencies I changed them a little bit (even from the current ones used in the requirements.txt file)
Regarding versioning: I was thinking on using github-release to prepare the releases (https://github.com/j0057/github-release). But up to now there was no real need, so nothing is existing. I have my "build" and test chain in a different repro: https://github.com/OzzieIsaacs/calibre-web-test/tree/master/build.
Where build means updating the translation files and generating an array of language names (for calibre to store the book language: "es" means "Spanish" in "English", "Spanisch" in "German", and so on)

decentral1se:
> Regarding the dependencies I changed them a little bit (even from the current ones used in the requirements.txt file)

Should I perhaps remove them and we focus on the `setup.cfg` now? I could also update the documentation.

Issue #933, opened 2019-06-07
[Docker] Allow "Reconnect to Calibre DB" REST call unauthenticated


yannouchou:
There are a few db sync issues that seem to be resolved by clicking "Reconnect to Calibre DB" (in my case, the updated books appear but not the covers after new books are added to Calibre - running Calibre-web in docker accessing the db over an NFS-mounted drive).

See #595 and #727 

While it would be nice to find out the root cause for this, in the meantime a simple workaround would be to simply automate the "Reconnect to DB" action after new books are added.

This could be easy to automate by sending a REST call to calibre-web. There is already an endpoint for this at `/shutdown?parameter=2` but it requires admin authentication, which means we need our automatic reload script to manage authentication and proper handling of the credentials.

Since making this call has a very limited impact on the applciation, I would suggest to open a separate anonymous endpoint for it (e.g. `/reconnect`).

I should be able to send a PR for this if you're ok with the idea.

OzzieIsaacs:
The PR is merged and you can use it, as soon as the docker provider updated the calibre-web installation

PR #934, opened 2019-06-07
Added public /reconnect endpoint

yannouchou:
Should address #933

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -1884,6 +1884,12 @@ def shutdown():
             return json.dumps({})
         abort(404)
 
+@app.route("/reconnect")
+def reconnect():
+    db.session.close()
+    db.engine.dispose()
+    db.setup_db()
+    return json.dumps({})
 
 @app.route("/search", methods=["GET"])
 @login_required_if_no_ano


yannouchou:
Hello there @OzzieIsaacs !
Any thoughts about this suggestion?

OzzieIsaacs:
Sorry for the late reply.
I have one question: How do you think users can use this "public api"? 
Shall they enter the url in their browser and wait for an empty reply, or do you want to integrate this into some sort of srcipt? 
I think this is not very comfortable for "normal" non programming users. (Maybe better than nothing, but not comfortable)

yannouchou:
In mind mind the idea was really mostly to support automation by scripts, typically to be executed after each new book addition in calibre.
End users already have the "reconnect" button in the UI, and for end users it's not going to be an issue to have to log in, whereas currently the requirement for being logged in makes scripting very difficult.

I have a working one-liner script but it was not that easy to figure out, only works with recent curl versions (older versions have a bug sending cookies to single-name domains), and of course requires managing and passing credentials which isn't easy to do if you really want to do it securely (and will fail with some complex password that break the form encoding):
```
curl "http://calibre-web/login" -d "username=${CALIBRE_ADMIN_USER}&password=${CALIBRE_ADMIN_PASS}&next=%2Fshutdown%3Fparameter%3D2" -L -b dummy
```

Whereas this could just be:
`curl "http://calibre-web/reconnect"`

Of course, this is not very elegant as it's still a workaround, but as you said it's probably better than nothing...

OzzieIsaacs:
I thought a bit about it, I think it's better to trigger the reconnect at certain events (book uploaded, ..?) from inside calibre-web in case of some environment variable is set. This environment variable could be set by the docker container and so the reconnect only happens for docker users.

yannouchou:
Hmm the problem is that many library updates happen outside of Calibre-web (e.g. from Calibre or a file sync tool). So we still need a way to externally notify Calibre-web and tell it "hey, please refresh"...
I however fully support the idea of using an env variable to enable/disable the new behaviour...

pwr:
An endpoint is not necessary, reloading can be triggered internally by just watching `metadata.db` for changes (modification time).
But this is not quite compatible with modifying the database from within CW itself, either with uploads, metadata editing or conversions. If the file is being sync'ed externally as well, getting conflicts is just a matter of time...
Without a global configuration flag to disable all these features in Calibre-Web, sync'ing the calibre database is just unsafe.

yannouchou:
Definitely there is no conflict-proof way of doing file syncing, but this is not the problem here.

The problem I, and I suspect many other users are facing, is simply that whenever an external update is made to the library, CW struggles to pick it up properly (e.g in my case new books will show up but not their cover image).

What I'm suggesting here with this API is definitely not a fix but a trivial workaround that I find good enough.

Continuously polling for the last file update would also work but might be an I/O overkill (in my case I know exactly when updates are made so I can trigger a reconnect just when it's needed), and I suspect not particularly easy to implement.

benhoh:
I'm really eager for this public endpoint. My issues with needing database reconnection are exactly the same as yannouchou's — I have Calibre-Web running on from a remote Calibre database that's a mirror of my laptop's Calibre setup. For me, calibre-web is a way to access my Calibre library from a reading device when my laptop's closed. This is why having the discretion to manage this external to calibre-web would be awesome.

Videl:
Same here, using Syncthing to synchronize the database.

MinchinWeb:
Is this documented anywhere? I think this might be the only place with is mentioned...

PR #956, opened 2019-07-01
Update LDAP management

Krakinou:
Hi,
I've  reviewed the changes and updated some stuff, mainly flask_simpleldap does not handle the exception very well, so they were not catched by your code, I updatyed it using the string of the exception

However, I encounter an issue on the change you've done in opds.py : 

````
File "/home/nicolas/calibre-web/cps/opds.py", line 49, in decorated
    return ldap.ldap.basic_auth_required(*args, **kwargs)
TypeError: basic_auth_required() takes exactly 2 arguments (1 given)
````

I don't understand why the args are not transferred to the basic_auth_required function (I'm still learning python, so I may have miss smthg obvious...)


OzzieIsaacs:
I have to setup a server (again) for using LDAP, to check it.
ldap.basic_auth_required needs the function to decorate as parameter, so a logical(?) approach would be to give "f" as parameter like in the example below.

``
def requires_basic_auth_if_no_ano(f):
    @wraps(f)
    def decorated(*args, **kwargs):
        auth = request.authorization
        if config.config_anonbrowse != 1:
            if not auth or not check_auth(auth.username, auth.password):
                return authenticate()
        return f(*args, **kwargs)
    if config.config_login_type == 1 and ldap_support:
        return ldap1.ldap.basic_auth_required(f)
    return decorated
``

Krakinou:
Ok, I got the issue.
Module are not imported in the correct sequence :  opds is initialized by cps.py while ldap is not yet, which cause issue in opds.py at server start up.
Will look into it and update the pull request

OzzieIsaacs:
I fixed it, the LDAP object is created upon first import and later on initialized

OzzieIsaacs:
I merged the code into develop and back to master, but I deactivated the function. From my point of view there are some open issues/topics:
How do we recover from wrong settings (from my experiments it's easier to do it wrong than to do it right)?
My suggestion would be add a test-config button to the ldap settings path. This would also guarantee that a admin user is existing for ldap login
Isn't ldap/ldaps related to tls/ssl config? So the questions is: Is it really necessary to show this setting to the user, can't we take the ssl/tls setting to determine if it's ldaps or ldap?
If we have ssl or tls enabled don't we need to provide a certificate for both cases? ( I'm ldap beginner, it took me 5 hours to get it basically work, I hope you know the answer)
I think it would be a good idea to check if the string entered in the user filter field is valid (contains "%s"), and give user feedback if it's not entered correctly
Final question: If we use ldap for authentication all passwords entered in calibre-web are obsolete, shouldn't we hide the password field for this case?

PR #958, opened 2019-07-07
Moved config class into separate file.

pwr:
Moved Goodreads and LDAP services into separate package; Gdrive and OAuth to follow.
Various other small fixes.


OzzieIsaacs:
Looks not so bad:
grafik
I'll try to fix it in the next week (I also introduced some of the errors with my last changes)

pwr:
How do you get these reports? I could make use of them before comitting :).

OzzieIsaacs:
I have my on test and build repository located at: https://github.com/OzzieIsaacs/calibre-web-test/tree/development
I wrote a short readme for getting started. The tests are done with unit tests and selenium. In the beginning I had the very clever idea to do system tests, as most of the function calls return a html-page  over the server. The outcome now is that it takes nearly 50mins to let all the tests run. And as I hate creating the tests, as it's such a boring job, the coding is not very good, there is lot of duplicate code and many errors aren't properly handled. 
BUT: It does a good job in finding bugs (encoding issues, error 500 pages have dropped significantly since I'm doing the tests).

PR #980, opened 2019-07-27
Add 2 modules to reqirements.txt

radoslawkierznowski:
Launching Calibre-Web on Windows displayed messages about missing modules.


OzzieIsaacs:
This modules are intentionally not in the requirements file included, as you only need them for Special features, they are Part of optional-requirements

PR #996, opened 2019-07-31
Update layout.html

stfrbrntyu:
Some minor visual tweaks in the top bar, so that the icons and the words have a space between them.

### DIFF
--- a/cps/static/css/caliBlur_override.css
+++ b/cps/static/css/caliBlur_override.css
@@ -1,0 +1,17 @@
+body.serieslist.grid-view div.container-fluid>div>div.col-sm-10:before{
+    display: none;
+}
+
+.cover .badge{
+    position: absolute;
+    top: 0;
+    right: 0;
+    background-color: #cc7b19;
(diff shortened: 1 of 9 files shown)


stfrbrntyu:
I might be doing something wrong here ;)
I only wanted to commit:
https://github.com/janeczku/calibre-web/pull/996/commits/f1f91469455e71dea3164592d924ae24af6a5ce0

OzzieIsaacs:
Shouldn't we do this via css?
In my browser it looks fine.
grafik

stfrbrntyu:
In the layout template it is also done with a space in front of "Advanced search", so I thought also to set it in front of Tasks, Admin, admin and Logout as it looks better when there is a space between the icon and the text. But I guess it's just nitpicking. I agree with css would be better.

`<span class="hidden-sm"> {{_('Advanced Search')}}</span>`

After the change:
image

OzzieIsaacs:
I'll remove it in front of Advanced search

PR #1014, opened 2019-08-19
Add automatic epub to kepub conversion using kepubify

vi-n:
**Intro**
Kobo e-readers use a specific epub variant that allows more advanced capabilities like pages left in this chapter, length of different chapters and faster page turns. 

When using Calibre I transfer files to my Kobo e-reader with the KoboTouchExtended plugin installed. This plugin automatically converts epub's to the so called "_kepubs_" that allow me to use the advanced Kobo features. In my kobo they are shown as being of the "kobo epub" type.

**Problem**
I would like to be able to use calibre-web on the experimental browser on the Kobo to directly download the books to my device without having to connect to a computer. I can do that with the current calibre-web but I lose the "kobo epub" functionality.

Calibre's convert utility that is already integrated in calibre-web is not able to convert from epub to _kepub_.

**Proposed solution**
There is an open source go utility called Kepubify which makes it very easy to convert epub's to _kepub's_: https://github.com/geek1011/kepubify. I propose to enable the user to integrate this into calibre-web in a similar way to the convert utility and then automatically using it when downloading a book on a device with a "Kobo" useragent. 

**Note**
The code in this pull request is not yet final, but it is working. I want to see if such change would be welcome in calibre web before putting the final touches on the change. If so, I will finish up this change and update this pull request. I chose to submit a pull request instead of a ticket to outline the change as I see it.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -303,6 +303,10 @@ def _configuration_update_helper():
     _config_string("config_calibre")
     _config_string("config_converterpath")
 
+    _config_checkbox_int("config_automatic_kepub")
+    _config_string("config_kepubify_path")
+    _config_string("config_kepub_cache_dir")
+
     if _config_int("config_login_type"):
         reboot_required |= config.config_login_type != constants.LOGIN_STANDARD
 
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -104,6 +104,10 @@ class _Settings(_Base):
     config_calibre = Column(String)
     config_rarfile_location = Column(String)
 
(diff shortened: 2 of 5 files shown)


Dinth:
Hi @vi-n im not sure if this was merged? Im also waiting for kepub support in calibre-web.

OzzieIsaacs:
It was not merged. I can't remember if it was based on a discussion in another issue and I denied merging it. But meanwhile the world has changed we are having the kobo sync protocol and therefore an relevant usecase for converting to kepub, and I'm also willing to implement the feature of sending emails in other formats the mobi, which also fits a little, little bit to this PR.

PR #1052, opened 2019-10-01
Add the possibility to change the username (#936)

### LINKED ISSUE #936: [Enhancement] Allow change of username without deleting and recreating user
Change user's username can only be changed if user is deleted and recreate it with new name.
Proposal:
It should be possible to change the users name (check if other user with same name is already present)

zelazna:
close #936

### DIFF
--- a/cps/templates/user_edit.html
+++ b/cps/templates/user_edit.html
@@ -3,7 +3,7 @@
 <div class="discover">
   <h1>{{title}}</h1>
   <form role="form" method="POST" autocomplete="off">
-    {% if g.user and g.user.role_admin() and new_user %}
+    {% if g.user or g.user.role_admin() or new_user %}
     <div class="form-group required">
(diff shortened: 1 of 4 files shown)


cps/templates/user_edit.html
```diff
@@ -3,7 +3,7 @@
 <div class="discover">
   <h1>{{title}}</h1>
   <form role="form" method="POST" autocomplete="off">
-    {% if g.user and g.user.role_admin() and new_user %}
+    {% if g.user or g.user.role_admin() or new_user %}
```
zelazna: not sure it's still usefull

OzzieIsaacs:
Thanks for the contribution

I changed the PR a little bit: Only admins can change the username and the anonymous username can't be chanded (there a some pieces of code which rely on the  name Guest for anonymous user).

PR #1107, opened 2019-12-13
Login via header sent from proxy

### LINKED ISSUE #1105: Forward auth/SSO support
My use case has this app sitting behind a reverse proxy with a forward auth mechanism that provides SSO for an entire domain. It would be great to be able to login via the username (passed by header attached to the request) returned by the SSO service.

aroberts:
This PR enables login via the contents of a header set by some upstream proxy.

If you're using this feature, it's assumed that there's no access to the calibre-web service without going through the proxy responsible for setting the headers, as it allows for unchallenged login (i.e., it assumes the login challenge happens upstream, in some SSO flow).

Fixes #1105

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -347,6 +347,10 @@ def _configuration_update_helper():
 
     _config_int("config_updatechannel")
 
+    # Reverse proxy login configuration
+    _config_checkbox("config_allow_reverse_proxy_header_login")
+    _config_string("config_reverse_proxy_login_header_name")
+
     # GitHub OAuth configuration
     if config.config_login_type == constants.LOGIN_OAUTH:
         active_oauths = 0
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -104,6 +104,9 @@ class _Settings(_Base):
     config_calibre = Column(String)
     config_rarfile_location = Column(String)
 
+    config_allow_reverse_proxy_header_login = Column(Boolean, default=False)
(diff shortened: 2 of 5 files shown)


OzzieIsaacs:
I've merged the commit. It would be great if you could provide a few sentences for the wiki how to use the feature.

aroberts:
I added a short section to the "Reverse Proxy" page.

rg9400:
I appreciate this feature a ton, +1. It seems to only work with users already set up in Calibre Web though, so if a new username is passed via the header, it won't create a login for them. As such, I can understand the reasons for not doing it, but I was able to get it to work in say Grafana to create a new user if they didn't exist before. That way, new users create their login upstream at my domain (via Organizr), and then there is seamless SSO into Calibre Web rather than having to register them again.

aroberts:
I see this as an accompaniment to the eventual LDAP bridge, which should either query ldap instead of the local db for users, or create users on some interval based on an ldap query. I don't see a good way for a header with a username to create users - what would the rest of the settings be? It doesn't feel right to assume defaults.

rg9400:
Well, there are default settings for new users already defined in the settings, but the main ones would be Email/Kindle Email, which I might be able to pass on from upstream as well. Password is unimportant with SSO. Essentially, the logic between the LDAP integration you are mentioning and my thought process is the same. Users are created and maintained at some other point, either LDAP or upstream at Organizr for example. Though for now, I can just manually duplicate my upstream users into CW and it should work for my usecase.

Issue #837, opened 2019-03-10
[Development] LDAP with Wrong DN and/or URL login is not possible anymore


OzzieIsaacs:
If wrong DN or LDAP url infos are entered it's not possible to enter calibre-web anymore, fallback routine or check during entering shall be implemented

Issue #263, opened 2017-08-04
Please add LDAP and/or AD auth


ruben-herold:
hi,

it would be nice if I could use ldap or ActiveDirectory for user athentication

ruben-herold:
https://code.tutsplus.com/tutorials/flask-authentication-with-ldap--cms-23101 example

ThorstenS-linux:
this feature would be really helpfull for our school-servers.

nocylah:
PAM auth would also be useful.

nstephenh:
I'd also like to request LDAP, although I don't really use accounts with this program it would be nice to have the option.

artiomn:
I need LDAP too.

lucas-dclrcq:
Hi, I'm currently working on an external authentication scheme for Calibre web and I have a question : are people more interested in an implementation of LDAP authentication or OpenID ?

bodybybuddha:
@lucas-dclrcq it looks like the biggest win is for LDAP, as it's mentioned more often then OpenID.  Of course it would be awesome to get them all!

@OzzieIsaacs also has mentioned in previous posts that he was hoping to do a plug in type system.  Maybe coordinate with him to ensure everyone is on the right path? Might have an easier time in getting the feature merged into the main fork.

Krakinou:
Hi,
Any update on this subject?
I would be very interested by LDAP too

Gorian:
Hello - I'd also like LDAP / Active Directory! I'm integrating everything on my network with Active Directory, and that would seel the deal for using this for hosting ebooks

Krakinou:
Tired of waiting :)
#769

centuryx476:
This would be an awesome feature.
Any update on it ?

BinsonBuzz:
+1 from me for LDAP

OzzieIsaacs:
LDAP is available in the newest master commit version

PR #1120, opened 2019-12-28
LDAP Import Users + Usability/Bug Fixes

ground7:
Hey there, I've been wanting to use this functionality for awhile as it's the last part of my home server that isn't integrated with LDAP. I attempted enabling and using the existing code and got it to work, but I ran into a couple of problems. I started to fix them as I came across them, and ending up adding additional functionality along the way. I'd like to address the main questions you asked in #956 so that we can get this enabled for more calibre-web users. #263 

The main thing I added was an LDAP User Import which replaces the create user button when selecting LDAP as the login type. It works by querying an LDAP group (everyone in the supplied ldap group is added), so there's a few extra parameters that are needed. It'll pull in username, password, email, (if there's a secondary email, it'll pull it in as a kindle email), and set the permissions/views to the defaults. This works well enough as a test, since it wont import if your settings aren't configured properly, and it'll query your LDAP database, so you'll be able to see the ldap search results in your ldap log files in order to debug your configuration.

The flow I've been using for setup is to ignore the ldap settings on the initial setup and log in with the default admin, and then tweak them from there. This better prevents you from locking yourself out in case you import an admin credential that you don't know (unlikely but still) and lets you test/import/tweak user settings. I've hidden the ldap option on initial setup.

If ldap login fails, it now tries basic auth before throwing ldap errors, so even if you swap to ldap login and then logout with a bad config, you can still log into the gui with the admin user. #837 

As for ssl/tls/ldaps/cert options, I think its best to leave these as there may be people with dated ldap configs that might need them. I'm not an ldap expert myself, but from what I understand this is the gist of how these options work:

ldap - doesn't restrict the port, defaults to 389
ldaps - must be a secure port (636)
ssl - outdated option that requires cert and will establish a secure connection using ssl
tls - similar to ssl with newer tls standard though still the dated way to configure your ldap database
cert is needed for those options

I use starttls which is an ldapv3 option that creates a secure connection with the ldap database after an initial query (that doesn't expose any unencrypted passwords if you set it up properly) I also use port 636 so my settings look like this:
ldaps
no ssl or tls or cert
If I check the additional options, it'll establish a tls connection with starttls and then error upon trying to encrypt it a second time with either ssl or tls. So basically if you are using an ldap database with a recommended config, then you won't use ssl or tls or cert options since you should be using starttls. That being said, I'm sure some people have their databases configured to use those options, so it's best to leave the options there.

Some other stuff I changed due to bugs:
The ldap database admin username field was unnecessarily restricted. It autofilled some information from other fields instead of letting you configure it separately. This doesn't work for openldap defaults, since you typically want to search for uid for your users and use cn=admin for your admin login. I removed the restrictions, and put various openldap examples in for the defaults to help people out. I haven't tested this with AD, but there's no reason it won't work so long as you know how to filter your options for AD.
Password doesn't autofill and must be reentered on changing config. I ran into a bug where the encrypted version of the password was being filled in, and I wasn't sure how to fill in the decrypted one if that's possible. In addition, after the first configuration, calibre-web must be restarted for ldap settings to register because of how the settings are being initialized. Not sure how you want to handle this. Maybe just have an alert saying restart required or something.

I didn't add much real time feedback to the forms as I was having some difficultly getting it to work when I did, but I figured some prefilled examples along with the import button should be adequate for an initial release. I don't believe it's possible to lock yourself out of an admin account now either unless you successfully import an ldap user named admin which you don't know the password to.

Let me know what else I can do to get my changes into a release.

### DIFF
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -37,6 +37,8 @@ _Base = declarative_base()
 class _Settings(_Base):
     __tablename__ = 'settings'
 
+    config_is_initial = Column(Boolean, default=True)
+    
     id = Column(Integer, primary_key=True)
     mail_server = Column(String, default='mail.example.org')
     mail_port = Column(Integer, default=25)
@@ -86,18 +88,21 @@ class _Settings(_Base):
 
     # config_oauth_provider = Column(Integer)
 
-    config_ldap_provider_url = Column(String, default='localhost')
+    config_ldap_provider_url = Column(String, default='example.org')
     config_ldap_port = Column(SmallInteger, default=389)
     config_ldap_schema = Column(String, default='ldap')
-    config_ldap_serv_username = Column(String)
+    config_ldap_serv_username = Column(String, default='cn=admin,dc=example,dc=org')
     config_ldap_serv_password = Column(String)
     config_ldap_use_ssl = Column(Boolean, default=False)
     config_ldap_use_tls = Column(Boolean, default=False)
     config_ldap_require_cert = Column(Boolean, default=False)
     config_ldap_cert_path = Column(String)
-    config_ldap_dn = Column(String)
-    config_ldap_user_object = Column(String)
-    config_ldap_openldap = Column(Boolean, default=False)
+    config_ldap_dn = Column(String, default='dc=example,dc=org')
+    config_ldap_user_object = Column(String, default='uid=%s')
+    config_ldap_openldap = Column(Boolean, default=True)
+    config_ldap_group_object_filter = Column(String, default='(&(objectclass=posixGroup)(cn=%s))')
+    config_ldap_group_members_field = Column(String, default='memberUid')
+    config_ldap_group_name = Column(String, default='calibreweb')
 
     config_ebookconverter = Column(Integer, default=0)
     config_converterpath = Column(String)
--- a/cps/web.py
+++ b/cps/web.py
@@ -54,7 +54,7 @@ from .pagination import Pagination
 from .redirect import redirect_back
 
 feature_support = {
-        'ldap': False, # bool(services.ldap),
+        'ldap': bool(services.ldap),
         'goodreads': bool(services.goodreads_support)
     }
 
@@ -253,6 +253,29 @@ def before_request():
         return redirect(url_for('admin.basic_configuration'))
 
 
+@app.route('/import_ldap_users')
+def import_ldap_users():
+    new_users = services.ldap.get_group_members(config.config_ldap_group_name)
+    for username in new_users:
+        user_data = services.ldap.get_object_details(user=username, group=None, query_filter=None, dn_only=False)
+        content = ub.User()
+        content.nickname = username
+        content.password = username # dummy password which will be replaced by ldap one
+        content.email = user_data['mail'][0]
+        if (len(user_data['mail']) > 1):
+            content.kindle_mail = user_data['mail'][1]
+        content.role = config.config_default_role
+        content.sidebar_view = config.config_default_show
+        content.mature_content = bool(config.config_default_show & constants.MATURE_CONTENT)
+        ub.session.add(content)
+        try:
+            ub.session.commit()
+        except Exception as e:
+            log.warning("Failed to create LDAP user: %s - %s", username, e)
+            ub.session.rollback()
+    return ""
+
+    
 # ################################### data provider functions #########################################################
 
 
@@ -1155,10 +1178,12 @@ def login():
                 if user and check_password_hash(str(user.password), form['password']) and user.nickname != "Guest":
                     login_user(user, remember=True)
                     flash(_(u"You are now logged in as: '%(nickname)s'", nickname=user.nickname), category="success")
+                    config.config_is_initial = False
                     return redirect_back(url_for("web.index"))
                 else:
                     log.info('Login failed for user "%s" IP-adress: %s', form['username'], ipAdress)
                     flash(_(u"Wrong Username or Password"), category="error")
+    
     settings = config.get_mail_settings()
     mail_configured = bool(settings.get("mail_server", "mail.example.org") != "mail.example.org")
 
--- a/cps/web.py
+++ b/cps/web.py
@@ -1155,7 +1155,11 @@ def login():
                 flash(_(u"you are now logged in as: '%(nickname)s'", nickname=user.nickname),
                       category="success")
                 return redirect_back(url_for("web.index"))
-            if login_result is None:
+            elif user and check_password_hash(str(user.password), form['password']) and user.nickname != "Guest":
+                login_user(user, remember=True)
+                flash(_(u"You are now logged in as: '%(nickname)s'", nickname=user.nickname), category="success")
+                return redirect_back(url_for("web.index"))
+            elif login_result is None:
                 flash(_(u"Could not login. LDAP server down, please contact your administrator"), category="error")
             else:
                 ipAdress = request.headers.get('X-Forwarded-For', request.remote_addr)
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -44,7 +44,7 @@ from .gdriveutils import is_gdrive_ready, gdrive_support
 from .web import admin_required, render_title_template, before_request, unconfigured, login_required_if_no_ano
 
 feature_support = {
-        'ldap': False, # bool(services.ldap),
+        'ldap': bool(services.ldap),
         'goodreads': bool(services.goodreads_support)
     }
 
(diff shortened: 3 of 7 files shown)


Krakinou:
I've been offline for some time due to multiple personnal issues but I'm now back and ready to help. They were still a lot of issues in the existing LDAP implementation, as mentioned in #956 and your update seems to solve a lot of them.

I agree with you regarding tls/ssl option.
Regarding the restart of calibre-web, there is an variable called "reboot_required" in admin.py, for [example](https://github.com/ground7/calibre-web/blob/b782489a8c9a4d946bbca3ec102d26bdb3305603/cps/admin.py#L312)

I've been trying your version with my ldap settings and here are some feedback:
- I did not succeed in making work the "Import LDAP users" option: nothing seems to happen when I click the button, I don't get anything in the logs neither (debug level), I'm not sure the function is triggered at all... 
- Shouldn't we put this function just before login also? In this way, if a user newly created in the LDAP repository tries to log in calibreweb, he will be directly created without any additionnal action from the admin?
- If I click on "Reset User password", I get an error "maximum recursion depth exceeded" where it seems like we are stuck in a loop : I think this option should be disabled when LDAP is activated

OzzieIsaacs:
> If I click on "Reset User password", I get an error "maximum recursion depth exceeded" where it seems like we are stuck in a loop : I think this option should be disabled when LDAP is activate

This is a bug in the current master Branche. I will Fix it next week. I‘m currently with limited internet connection

ground7:
> * I did not succeed in making work the "Import LDAP users" option: nothing seems to happen when I click the button, I don't get anything in the logs neither (debug level), I'm not sure the function is triggered at all...

I personally use osixia's openldap docker container, so I just look at my ldap log to see how calibre-web is attempting searches. If you see nothing in your ldap debug log, then it might be because I switched how the bind login was configured to use the full dn back from how you had it. I agree that this needs some GUI/log feedback though now that you bring it up since people might not be able to view their LDAP logs.

If you see the queries attempted in your LDAP logs but not passing, you'll have to tweak the group settings. The import takes your group name and puts it in your group filter, then looks for members in that group using the member filter (defaults to memberUid for openldap). If you don't have a group with members in your ldap database, it won't work. Also, if there's already members in calibre-web with the email from ldap, it'll fail to import them.

> * Shouldn't we put this function just before login also? In this way, if a user newly created in the LDAP repository tries to log in calibreweb, he will be directly created without any additionnal action from the admin?

Yes, this seems like a good feature.

We could technically get the reset password to work too, but I use Nextcloud as a user frontend, and it already does this, so I just have my users reset their LDAP passwords through there.

Krakinou:
I misexplain : Clicking on the "Import LDAP User" does not trigger the function.
- No LDAP log
- No debug message in /import_ldap_user (I tried to add some debug messages)

cps/web.py
```diff
@@ -1155,10 +1182,12 @@ def login():
                 if user and check_password_hash(str(user.password), form['password']) and user.nickname != "Guest":
                     login_user(user, remember=True)
                     flash(_(u"You are now logged in as: '%(nickname)s'", nickname=user.nickname), category="success")
+                    config.config_is_initial = False
                     return redirect_back(url_for("web.index"))
                 else:
                     log.info('Login failed for user "%s" IP-adress: %s', form['username'], ipAdress)
                     flash(_(u"Wrong Username or Password"), category="error")
+
```
OzzieIsaacs: Codacy Issue found: [Trailing whitespace](https://app.codacy.com/manual/Ozzie-IsaacsOrganization/calibre-web/pullRequest?prid=4846307)

OzzieIsaacs:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Ozzie-IsaacsOrganization/calibre-web/pullRequest?prid=4846307&bid=16085443)

cps/config_sql.py
```diff
@@ -37,6 +37,8 @@
 class _Settings(_Base):
     __tablename__ = 'settings'
 
+    config_is_initial = Column(Boolean, default=True)
+
```
OzzieIsaacs: Codacy Issue found: [Trailing whitespace](https://app.codacy.com/manual/Ozzie-IsaacsOrganization/calibre-web/pullRequest?prid=4846307)

OzzieIsaacs:
The PR was missing some error handling which I added over the last few days:
- python3 fixes (storage of passwords, import user and email)
- password is no longer shown in UI same as Users password, as the stored password is encoded and doesn't reflect what is displayed, so changing one letter before would destroy password totally
- if logged in via fallback login, this is now displayed
- removed choice for Ldap/ldaps, TLS/SSL is selection, and only if encryption is choosen the certificate is displayed. Option to select certificate is removed (choosen yes if certificate is not None, otherwise no). 
- Feedback dialog on user import
- Errorhandling on user import (LDAP not available, empty response from server)
- Handled no email field in user import
- perform automatic reboot on LDAP changed options
- double import of same username is handled properly
- input validation check (%s has to be present) on LDAP user query and group query
- Additional handling of "groupofnames" system of group managment (the userlist in this case is not the username, but the query to get the username like ("uid=name,dn=example,dn=com" instead of "username")
- Handling of no password case given in note on flask simpleLDAP doc page (https://flask-simpleldap.readthedocs.io/en/latest/)
- Handling of case insensitive user names on import
-  check equal number of parenthesis during LDAP config

It was a lot of work, now it's done. I also added a section to the wiki reflecting my opinion how to configure. Would be great if you could check it, and also if the changes are working for you (I created some automatic tests in my testing repro, so it should work, only thing missing is certificate check)

Krakinou:
Great work!
I made some tests and the connection of the users is working. I did not try yet user import.
I find just a bug so far : 
- Set all parameters for LDAP connection
- Set something wrong in the parameters for user import (eg : no %s)
- Save: an error message is issued as expected
However, if you log out and try to log in again with a LDAP user, following message is issued;
```Traceback (most recent call last):
File "/home/nicolas/calibre-web/vendor/flask/app.py", line 2447, in wsgi_app
response = self.full_dispatch_request()
File "/home/nicolas/calibre-web/vendor/flask/app.py", line 1952, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/home/nicolas/calibre-web/vendor/flask/app.py", line 1821, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "/home/nicolas/calibre-web/vendor/flask/_compat.py", line 39, in reraise
raise value
File "/home/nicolas/calibre-web/vendor/flask/app.py", line 1950, in full_dispatch_request
rv = self.dispatch_request()
File "/home/nicolas/calibre-web/vendor/flask/app.py", line 1936, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
File "/home/nicolas/calibre-web/cps/web.py", line 1212, in login
login_result, error = services.ldap.bind_user(form['username'], form['password'])
File "/home/nicolas/calibre-web/cps/services/simpleldap.py", line 88, in bind_user
if _ldap.get_object_details(username):
File "/home/nicolas/calibre-web/vendor/flask_simpleldap/__init__.py", line 171, in get_object_details
fields = current_app.config['LDAP_USER_FIELDS']
KeyError: 'LDAP_USER_FIELDS'```

I will keep on testing tomorrow.
Also, I made some update on the documentation to improve readability and correct some typo.

OzzieIsaacs:
I could not reproduce your problem, but added the KeyError to the handled exceptions

ground7:
Hey thanks for carrying this over the finish line. I recently updated to the new version and everything is working for me except one thing - the opds server downloads when using ldap.

I can properly authenticate against the opds server, but downloads will fail.
In [this commit](https://github.com/janeczku/calibre-web/pull/1120/commits/b782489a8c9a4d946bbca3ec102d26bdb3305603), I did an improper fix to get it working on my instance, and I'm not sure what's required to get it working correctly.

OzzieIsaacs:
The decorater line you removed is for checking if the current user has the access right to download books.By removing it, everybody can download books. From my point of view there are 2 possibilitys:
1) The user really has no download rights
2) there is something wrong in the code and the download_required function is called without active user (I had this one in the past, was a bit ugly)
Could you please verify no 1 and if this is not the case, please open a issue (link to your commit would be enough) I will try to check it within the next couple of days

OzzieIsaacs:
And, this is indeed a testing gap. I have to add tests for the opds feed with ldap login, to make sure this stays working

ground7:
> 
> Could you please verify no 1 and if this is not the case, please open a issue (link to your commit would be enough) I will try to check it within the next couple of days
> 

I’m able to download fine through a browser, just not via opds. I’ll open a new issue like you suggested.

PR #1137, opened 2020-01-12
custom columns: add Float type support

garf1242:
- remove float datatype from cc_exceptions
- handle float datatype like int or bool
- change edit template to modify input step (0.01 precision, like
  in calibre UI)


garf1242:
Thanks!

PR #1138, opened 2020-01-12
caliBlur theme: fix custom column display that contain colons

garf1242:
caliBlur theme uses split method to separate custom columns key and value with the colon separator.
However, when the value contains colons, it splits too many times, which ends up to a truncated output.

This quickly fixes this behavior (split, shift, then join) to get the correct output.


garf1242:
Thanks!

PR #1150, opened 2020-01-20
Update messages.po

nykul:


OzzieIsaacs:
The first new translation is missing a %, could you please add it

nykul:
> The first new translation is missing a %, could you please add it

done, sorry missed it

PR #1164, opened 2020-01-26
Add book Archive (i.e Trash) to CalibreWeb

shavitmichael:
Add support for per-user book Archiving (i.e trash):
- Archived books are hidden from the user, and are only visible from the new "Archived Books" tab.
- Archiving a book from the Web UI will remove it from any paired Kobo device on their next sync.
- Deleting a book from a paired Kobo device moves it to the user's Archive.

Manually tested the following interactions:
- Archived a book in the web UI and verified that it got removed from the Kobo device.
- Restored a book in the web UI and verified that it got added back to the Kobo device.
- Archived a book from the Kobo device, and verified that it was archived in the web UI.
- Verified that archived books don't appear in web UI searches, or tabs other than the Archived tab.

(Note: One of the commits in this PR comes from #1163)


OzzieIsaacs:
Hi, I hadn't time to test it, I not not sure if I made it correct, now the ...read_status... was overwritten with the  ...archived...  status, was this right?

PR #1165, opened 2020-01-26
Fix for viewing error logfile in admin

ghighi3f:
Hello again.

The view logfiles feature works as expected for accesslog file, but for calibre-web.log, /admin/logfile returns always the "Stream output, can't be displayed" message.

Screenshot 2020-01-26 at 20 25 38

Template logviewer.html expects a log_enable parameter `{% if log_enable %}` , but in the render function it's totally missing.

Because the **View logfiles** button is always visible, and i couldn't find any setting that could disable the logfile view, i assumed that this parameter should be True as default.

Hope I'm clear. All the best


OzzieIsaacs:
Yes, it's indeed a problem. Caused by mismerge from develop branch. The log_enable variable content should be:
log_enable=bool(config.config_logfile != logger.LOG_TO_STDOUT),

ghighi3f:
Should be ok now

PR #1174, opened 2020-02-03
Update Simplified Chinese translation

xcffl:
The "Read" string here stands for "have read already" instead of "to read".


OzzieIsaacs:
Thanks for the translation

PR #1187, opened 2020-02-09
Fixes cover layout in user edit page

Kyosfonica:


OzzieIsaacs:
Could you please move the style to the main.css file. I'd prefer to have the html file free of css directives.
Thanks

Kyosfonica:
Sure, I moved it to style.css because it already contains a cover class.

OzzieIsaacs:
Thanks, a lot. 
I modified it a little bit, to make the covers smaller. So they fit in size to the other cover pictures.

PR #986, opened 2019-07-29
Add support for grid view

gpapin:
Not too sure if I should base this on develop or master so feel free to edit.
This is pretty much done but still need to address 2 issues

- [ ]  Fix the alphabetical alphabetical sorting asc and desc in the cover view
- [x] Figure out what `B,A <-> A B` does and fix it in the cover view
- [ ]  Fix tooltip alignment on large displays
 
Feedback and pointers welcome.

PR #1214, opened 2020-02-22
Improve read status visibility

celogeek:
This patch contains:
 - fixed for get_cc_columns (it was inverted)
 - auto filter custom columns Read if selected in settings but keep the advanced search possible! (it does not display the read custom header in the book details, but allow advance search on it because we don't filter it from everywhere)
 - factorization of the code for the list of readBookIds.
 - Display of read status on every template
 - Keep search term and focus on page reload
 - Order by "sort" (great when looking for a series)

Capture d’écran 2020-02-22 à 19 50 58

### DIFF
--- a/cps/templates/layout.html
+++ b/cps/templates/layout.html
@@ -42,7 +42,7 @@
           <form class="navbar-form navbar-left" role="search" action="{{url_for('web.search')}}" method="GET">
             <div class="form-group input-group input-group-sm">
               <label for="query" class="sr-only">{{_('Search')}}</label>
-              <input type="text" class="form-control" id="query" name="query" placeholder="{{_('Search Library')}}">
+              <input type="text" class="form-control" id="query" name="query" placeholder="{{_('Search Library')}}" value="{{searchterm}}">
               <span class="input-group-btn">
                 <button type="submit" id="query_submit" class="btn btn-default">{{_('Search')}}</button>
               </span>
@@ -226,6 +226,11 @@
             $("#btn-upload").change(function() {
                 $("#form-upload").submit();
             });
+            $(document).ready(function() {
+              var inp = $('#query').first()
+              var val = inp.val()
+              inp.val('').blur().focus().val(val)
+            })
         });
     </script>
     {% block js %}{% endblock %}
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -803,16 +803,20 @@ def get_search_results(term):
             func.lower(db.Books.title).ilike("%" + term + "%")
             )).all()
 
-def get_cc_columns():
+def get_cc_columns(filter_config_custom_read=False):
     tmpcc = db.session.query(db.Custom_Columns).filter(db.Custom_Columns.datatype.notin_(db.cc_exceptions)).all()
+    cc = []
+    r = None
     if config.config_columns_to_ignore:
-        cc = []
-        for col in tmpcc:
-            r = re.compile(config.config_columns_to_ignore)
-            if not r.match(col.name):
-                cc.append(col)
-    else:
-        cc = tmpcc
+        r = re.compile(config.config_columns_to_ignore)
+
+    for col in tmpcc:
+        if filter_config_custom_read and config.config_read_column and config.config_read_column == col.id:
+            continue
+        if r and r.match(col.label):
+            continue
+        cc.append(col)
+
     return cc
 
 def get_download_link(book_id, book_format):
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -855,6 +855,20 @@ def check_exists_book(authr,title):
             func.lower(db.Books.title).ilike("%" + title + "%")
             )).first()
 
+def get_readbooks_ids():
+    if not config.config_read_column:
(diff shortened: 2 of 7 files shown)


celogeek:
Hi,
@OzzieIsaacs @Rewerson can you take a look at my patches, and tell me if I need to change anything?

Do you think it is good to integrate?

OzzieIsaacs:
Sorry for the late reply, I was busy in other parts of the project. Today I had a look at it. I like the idea of having the read status visible on the main screen. There was a similar request upon integrating audiobooks (the musical glyphicon), I removed the icons there because I think the layout looks a bit unsteady/uneasy and creates additional lines of text beside the beautiful cover pictures. My idea would go something like this (the badge):
grafik
My idea would go towards just the "cross" or the "check" (red colored maybe, so have it visible, or maybe on a plain little circle -> we should start with an easy implementation and look what it looks like). 
I gave the upper left (or right corner) to the Grid series view, so you could use the lower left, or right (but this looks like to cause trouble) corner (for details see: #986)

celogeek:
Wow yeah, nice feature! So this #986 is already merged somewhere or it is still in progress?

I may rework on my patch after that to feet this new feature before we merge anything.

I will also do several pull requests to separated the features.

OzzieIsaacs:
The feature isn‘t meted yet, as there are some bugs in it

celogeek:
I have start splitting features from this patch, it will be easier to integrate.

The first one: https://github.com/janeczku/calibre-web/pull/1315

When you select a custom read column in config, you won't have to filter it anymore.
So that way, you can see only one read status in Show Book, and you can use it in Advanced search.

If you filter the Read field, it disappears from Advance Search, and that is a problem.

celogeek:
Second one:

https://github.com/janeczku/calibre-web/pull/1316

The search result is now sorted by "Books.sort". This way search and advance search, or series show is always display in sort order.

When I take a look at a series I haven't read it all yet, I want to see the series number in a right order in order to be able to read then move on to the next ...

celogeek:
The third one:
https://github.com/janeczku/calibre-web/pull/1317

This keep the value of the search bar, and also keep focus.

So you enter:

"bef", then enter, you see the result,
you can complete: "ore"
you search input look like: "before", and go on

It speed up quick lookup and readability

celogeek:
For the last one, with the read status, I may wait for the bubble to be integrated.

I will try to add a filter until I can display the status nicely.

Capture d’écran 2020-04-19 à 13 16 29

Something like series/unread/42

cps/templates/layout.html
```diff
@@ -226,6 +226,11 @@ <h4 class="modal-title" id="bookDetailsModalLabel">{{_('Book Details')}}</h4>
             $("#btn-upload").change(function() {
                 $("#form-upload").submit();
             });
+            $(document).ready(function() {
+              var inp = $('#query').first()
+              var val = inp.val()
+              inp.val('').blur().focus().val(val)
+            })
```
menasamuel, line 233: Hi @celogeek - thanks very much for your contributions to this project! I use it extensively at home. :) 

Regarding this change which was merged to master about a month ago, would you mind walking me through why the value was cleared, then focus on the search box was forced on the user and then the value re-set? I don't understand why it should clear and re-set the value and why focus? It feels a bit jarring to the user does it not?

OzzieIsaacs: Looks like suggested method 3 to get the cursor to the end of a focused text field:
https://subinsb.com/move-cursor-to-end-input-textarea-jquery/

celogeek: Yeah, I pick the trick from somewhere on the web.
The step to be able to have a field focus and be at the end of the current text is:

- clear
- blur (unfocus)
- focus
- set back the value

This way you are focus on the field, and your cursor is placed at the end of the input text.

menasamuel: Thanks very much for explaining. Would it be a good idea to only do this if the value is not blank? Given that this logic is run on every page regardless of whether or not there is a search term present, it causes mobile keyboards to jump up while the user is scrolling which can be quite jarring. When I noticed it start happening recently I thought I had somehow broken something in my library. Maybe something along the lines of: 

`if (val.trim()) {
  inp.val('').blur().focus().val(val);
}`

Just a thought - I am sorted for now as I've just commented out that code but it might be a good idea for the next PR. 

Thanks for being patient as well - this is the first time I've done anything other than lurk on any code repos :)

celogeek: I have put that:
https://github.com/janeczku/calibre-web/pull/1418

OzzieIsaacs:
Okay the bubbles are now merged to master branch:
grafik
So you could start the read unread bubble programming (I think this was what we agreed)

celogeek:
Hi, I'm working on it.
First, something is wrong with advanced search, working on it. I will restore read search there.

celogeek:
I start the read bubble here: https://github.com/janeczku/calibre-web/pull/1662
I still need to get the read status to activate the mark only on read.
I close this PR, we can move to this one.

PR #1218, opened 2020-02-23
Update italian message.po

ElQuimm:
Thank you for your job :-)


OzzieIsaacs:
Just for your information: you don‘t need to translate the strings starting with `#~`, this are old strings, which are no longer in use

OzzieIsaacs:
Thank you very much, Italian translation is complete again

PR #1229, opened 2020-02-26
Update Chinese translation

flashlab:
btw, [crowdin.com](https://crowdin.com/) is a good tool to host i18n work.


OzzieIsaacs:
Thanks for the translation.
If I would use crowdin for translation this would add another hurdles for the users to update the translation, another website theyhad to apply for, so currently I lease it as it is. But thanks for the tip I will keep it in mind.
Wish you and your family all the best and stay healthy.

PR #1238, opened 2020-03-01
Add sitemap support.

xcffl:
I closed the last pr by mistake. This one is tested. 

Let me know if there is documentation work needed. I would like to add some to the master branch since I observe active development on that branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ settings.yaml
 gdrive_credentials
 client_secrets.json
 
+.vscode/settings.json
+.vscode/launch.json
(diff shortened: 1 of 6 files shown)


OzzieIsaacs:
I thought a bit about this and I came to the conclusion that I will not merge it to Calibre-Web. Calibre-Web is made for sharing books among your family or within a company. 
In my opinions there is no use case for getting indexed by a search engine.  
Sorry

xcffl:
Although in some cases it's shared within the family, it could be used to illustrate educational books that is free to distribute. I still think it's a good idea to have this as an option.

PR #1173, opened 2020-02-03
English Language Updates

joshobriencom:
Changes to provide consistency for English language.

Issue #657, opened 2018-10-02
Email test says successful no matter what's entered into fields


silentArtifact:
Hello,

Running into an odd issue; no matter what information I put into any of the SMTP fields, I can hit save and test, and it says that settings are saved, and email successfully sent. No emails are sending, obviously.

Edit: I noticed that under tasks it does show that the email tests fail, just not in the notifications that pop up on "save and test".

bodybybuddha:
Sounds like we need to change the label to email was successfully queued.  Also I would like for the save action to take you immediately to the task screen - so one could see the result of the test.

silentArtifact:
Ahh, that explains a lot. I'll have to open another one on why the email isn't actually sending, but it sounds like this one is mostly closed. Thanks!

OzzieIsaacs:
I changed the text of the message to make it more clear what's going on. Bypassing the queue and sending it out directly make to much work for this problem

PR #1264, opened 2020-03-19
English Language Updates V2

joshobriencom:
Changes to provide consistency for English language V2.

This continues from Pull Request #1173 with even further English language changes. 
https://github.com/janeczku/calibre-web/pull/1173


cps/templates/modal_restriction.html
```diff
@@ -3,10 +3,10 @@
...
       <div class="modal-header">
-        <h4 class="modal-title hidden" id="h1">{{_('Select allowed/denied Tags')}}</h4>
-        <h4 class="modal-title hidden" id="h2">{{_('Select allowed/denied Custom Column values')}}</h4>
-        <h4 class="modal-title hidden" id="h3">{{_('Select allowed/denied Tags of user')}}</h4>
-        <h4 class="modal-title hidden" id="h4">{{_('Select allowed/denied Custom Column values of user')}}</h4>
+        <h4 class="modal-title hidden" id="h1">{{_('Select Allowed/Denied Tags')}}</h4>
+        <h4 class="modal-title hidden" id="h2">{{_('Select Allowed/Denied Custom Columns')}}</h4>
+        <h4 class="modal-title hidden" id="h3">{{_('Select Allowed/Denied Tags')}}</h4>
+        <h4 class="modal-title hidden" id="h4">{{_('Select Allowed/Denied Custom Columns')}}</h4>
       </div>
```
OzzieIsaacs, line 10: I'm not happy with all changes: In general we are talking about visibility restrictions. The 4 lines are selected based on the current location from where the dialog was called.  
Select allowed/denied Custom Column values -> In this dialog you can enter values which have to be matched to a certain content of a custom column and you don't enter names of custom columns, what your suggested text implies. Maybe there is a better word than "value" 
Select allowed/denied Tags of user and Select allowed/denied Custom Column values of user are used if an admin edits the visibility restrictions of a certain user in contrast to the other 2 options, where you edit restrictions for yourself or as template for all users. So from my point of view we can not remove the user part.

cps/templates/modal_restriction.html
```diff
@@ -21,7 +21,7 @@ <h4 class="modal-title hidden" id="h4">{{_('Select allowed/denied Custom Column
           </table>
           <form id="add_restriction" action="" method="POST">
           <div class="form-group required">
-            <label for="add_element">{{_('Add View Restriction')}}</label>
+            <label for="add_element">{{_('Add New Restriction')}}</label>
```
OzzieIsaacs, line 24: I wanted to express that the restrictions are for the View/visibility of books, maybe "Add New Visibility Restriction" is a good description?

cps/web.py
```diff
@@ -1057,13 +1057,13 @@ def send_to_kindle(book_id, book_format, convert):
         result = send_mail(book_id, book_format, convert, current_user.kindle_mail, config.config_calibre_dir,
                                   current_user.nickname)
         if result is None:
-            flash(_(u"Book successfully queued for sending to %(kindlemail)s", kindlemail=current_user.kindle_mail),
+            flash(_(u"Success! Selected book has been sent to your Kindle (%(kindlemail)s)", kindlemail=current_user.kindle_mail),
                   category="success")
```
OzzieIsaacs, line 1061: Not a good idea, there is a similar sentence in the program, where I already got negative feedback about (#657). Because it does what it says, the book is only queued and if there several other tasks are pending it takes some time to send the book. This should be expressed in the text

cps/templates/user_edit.html
```diff
@@ -80,8 +80,8 @@ <h1>{{title}}</h1>
           <label for="Show_detail_random">{{_('Show Random Books')}}</label>
       </div>
       {% if ( g.user and g.user.role_admin() and not new_user ) %}
-      <a href="#" id="get_user_tags" class="btn btn-default" data-toggle="modal" data-target="#restrictModal">{{_('Add allowed/denied Tags')}}</a>
-      <a href="#" id="get_user_column_values" class="btn btn-default" data-toggle="modal" data-target="#restrictModal">{{_('Add allowed/denied custom column values')}}</a>
+      <a href="#" id="get_user_tags" class="btn btn-default" data-toggle="modal" data-target="#restrictModal">{{_('Edit Tag Restrictions')}}</a>
+      <a href="#" id="get_user_column_values" class="btn btn-default" data-toggle="modal" data-target="#restrictModal">{{_('Edit Custom Column Restrictions')}}</a>
```
OzzieIsaacs, line 84: See my comments on the other "View Restrictions" texts. I'm fine with Edit instead of add, you are right this is what it does. Would "Edit Tag Visibilities" something plausible? I have no good idea for the Custom Column one

OzzieIsaacs:
I've merged your changes and left the texts I gave comments about. I'm happy to hear from you, If you have some idea to change the texts

PR #1266, opened 2020-03-22
[Kobo] Add Shelf/Collection support.

shavitmichael:
Implements the v1/library/tags set of endpoints to sync CalibreWeb
shelves with the Kobo device.

Drive-by: Refactors shelf.py to consolidate how user permissions are checked.
Drive-by: Fix issue with the sync endpoint that arrises when a book is
hard-deleted.

Tested: I've been using this for the past week without noticable issues, although I'll admit I don't use Shelves as much now that Kobo has added the Series tab.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -407,7 +407,7 @@ def HandleTagCreate():
         log.debug("Received malformed v1/library/tags request.")
         abort(400, description="Malformed tags POST request. Data is missing 'Name' or 'Items' field")
 
-    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.name) == name, ub.Shelf.user_id ==
+    shelf = ub.session.query(ub.Shelf).filter(ub.Shelf.name == name, ub.Shelf.user_id ==
                                               current_user.id).one_or_none()
     if shelf and not shelf_lib.check_shelf_edit_permissions(shelf):
         abort(401, description="User is unauthaurized to edit shelf.")
@@ -426,7 +426,7 @@ def HandleTagCreate():
 
 @kobo.route("/v1/library/tags/<tag_id>", methods=["DELETE", "PUT"])
 def HandleTagUpdate(tag_id):
-    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.uuid) == tag_id,
+    shelf = ub.session.query(ub.Shelf).filter(ub.Shelf.uuid == tag_id,
                                               ub.Shelf.user_id == current_user.id).one_or_none()
     if not shelf:
         log.debug("Received Kobo tag update request on a collection unknown to CalibreWeb")
@@ -487,7 +487,7 @@ def HandleTagAddItem(tag_id):
         log.debug("Received malformed v1/library/tags/<tag_id>/items/delete request.")
         abort(400, description="Malformed tags POST request. Data is missing 'Items' field")
 
-    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.uuid) == tag_id,
+    shelf = ub.session.query(ub.Shelf).filter(ub.Shelf.uuid == tag_id,
                                               ub.Shelf.user_id == current_user.id).one_or_none()
     if not shelf:
         log.debug("Received Kobo request on a collection unknown to CalibreWeb")
@@ -498,7 +498,7 @@ def HandleTagAddItem(tag_id):
 
     items_unknown_to_calibre = add_items_to_shelf(items, shelf)
     if items_unknown_to_calibre:
-        log.debug("Received request to add an unknown book to a collecition. Silently ignoring item.")
+        log.debug("Received request to add an unknown book to a collection. Silently ignoring item.")
 
     ub.session.merge(shelf)
     ub.session.commit()
@@ -600,7 +600,7 @@ def create_kobo_tag(shelf):
         book = db.session.query(db.Books).filter(db.Books.id == book_shelf.book_id).one_or_none()
         if not book:
             log.info(u"Book (id: %s) in BookShelf (id: %s) not found in book database",  book_shelf.book_id, shelf.id)
-            return None
+            continue
         tag["Items"].append(
             {
                 "RevisionId": book.uuid,
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -17,9 +17,11 @@
 #  You should have received a copy of the GNU General Public License
 #  along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-import datetime
-import sys
 import base64
+import datetime
+import itertools
+import json
+import sys
 import os
 import uuid
 from time import gmtime, strftime
@@ -45,7 +47,7 @@ from sqlalchemy import func
 from sqlalchemy.sql.expression import and_, or_
 import requests
 
-from . import config, logger, kobo_auth, db, helper, ub
+from . import config, logger, kobo_auth, db, helper, shelf as shelf_lib, ub
 from .services import SyncToken as SyncToken
 from .web import download_required
 from .kobo_auth import requires_kobo_auth
@@ -119,6 +121,7 @@ def redirect_or_proxy_request():
 
 def convert_to_kobo_timestamp_string(timestamp):
     return timestamp.strftime("%Y-%m-%dT%H:%M:%SZ")
+
 
 @kobo.route("/v1/library/sync")
 @requires_kobo_auth
@@ -203,24 +206,23 @@ def HandleSyncRequest():
                      ub.KoboReadingState.user_id == current_user.id,
                      ub.KoboReadingState.book_id.notin_(reading_states_in_new_entitlements))))
     for kobo_reading_state in changed_reading_states.all():
-        book = db.session.query(db.Books).filter(db.Books.id == kobo_reading_state.book_id).one()
-        sync_results.append({
-            "ChangedReadingState": {
-                "ReadingState": get_kobo_reading_state_response(book, kobo_reading_state)
-            }
-        })
-        new_reading_state_last_modified = max(new_reading_state_last_modified, kobo_reading_state.last_modified)
+        book = db.session.query(db.Books).filter(db.Books.id == kobo_reading_state.book_id).one_or_none()
+        if book:
+            sync_results.append({
+                "ChangedReadingState": {
+                    "ReadingState": get_kobo_reading_state_response(book, kobo_reading_state)
+                }
+            })
+            new_reading_state_last_modified = max(new_reading_state_last_modified, kobo_reading_state.last_modified)
+
+    sync_shelves(sync_token, sync_results)
 
     sync_token.books_last_created = new_books_last_created
     sync_token.books_last_modified = new_books_last_modified
     sync_token.archive_last_modified = new_archived_last_modified
     sync_token.reading_state_last_modified = new_reading_state_last_modified
 
-    if config.config_kobo_proxy:
-        return generate_sync_response(request, sync_token, sync_results)
-
-    return make_response(jsonify(sync_results))
-    # Missing feature: Detect server-side book deletions.
+    return generate_sync_response(request, sync_token, sync_results)
 
 
 def generate_sync_response(request, sync_token, sync_results):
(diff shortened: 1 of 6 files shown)


shavitmichael:
Bump on this PR?
This is the one remaining large features that I intend to implement.

OzzieIsaacs:
I haven't forgotten you. I'm currently working on the LDAP integration, wrote some test, and I'm almost finished, I hope I can start working on this one end of the week. But it will take some time to merge. I'll write some tests(https://github.com/OzzieIsaacs/calibre-web-test), to ensure everything works well and I will start a little documentation in the wiki.

shavitmichael:
No worries. I actually just added a first revision for some [documentation](https://github.com/janeczku/calibre-web/wiki/Kobo-integration) to the wiki a few hours ago :)

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+    name, items = None, None
+    try:
+        name = shelf_request["Name"]
+        items = shelf_request["Items"]
+    except KeyError:
+        log.debug("Received malformed v1/library/tags request.")
+        abort(400, description="Malformed tags POST request. Data is missing 'Name' or 'Items' field")
+
+    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.name) == name, ub.Shelf.user_id ==
+                                              current_user.id).one_or_none()
```
OzzieIsaacs, line 411: I think the parthesis around `and_(ub.Shelf.name)` make no sense, should be around whole filter request?

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+        log.debug("Received request to add unknown books to a collection. Silently ignoring items.")
+    ub.session.commit()
+
+    return make_response(jsonify(str(shelf.uuid)), 201)
+
+
+@kobo.route("/v1/library/tags/<tag_id>", methods=["DELETE", "PUT"])
+def HandleTagUpdate(tag_id):
+    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.uuid) == tag_id,
+                                              ub.Shelf.user_id == current_user.id).one_or_none()
```
OzzieIsaacs, line 430: also suspicious parenthesis around `and_(ub.Shelf.name)`

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+    tag_request = request.json
+    items = None
+    try:
+        items = tag_request["Items"]
+    except KeyError:
+        log.debug("Received malformed v1/library/tags/<tag_id>/items/delete request.")
+        abort(400, description="Malformed tags POST request. Data is missing 'Items' field")
+
+    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.uuid) == tag_id,
+                                              ub.Shelf.user_id == current_user.id).one_or_none()
```
OzzieIsaacs, line 491: parenthesis like above

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+        log.debug("Received Kobo request on a collection unknown to CalibreWeb")
+        abort(404, description="Collection isn't known to CalibreWeb")
+
+    if not shelf_lib.check_shelf_edit_permissions(shelf):
+        abort(401, description="User is unauthaurized to edit shelf.")
+
+    items_unknown_to_calibre = add_items_to_shelf(items, shelf)
+    if items_unknown_to_calibre:
+        log.debug("Received request to add an unknown book to a collecition. Silently ignoring item.")
+
```
OzzieIsaacs, line 502: typo: collecition -> collection

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+    items = None
+    try:
+        items = tag_request["Items"]
+    except KeyError:
+        log.debug("Received malformed v1/library/tags/<tag_id>/items/delete request.")
+        abort(400, description="Malformed tags POST request. Data is missing 'Items' field")
+
+    shelf = ub.session.query(ub.Shelf).filter(ub.Shelf.uuid == tag_id,
+                                              ub.Shelf.user_id == current_user.id).one_or_none()
+    if not shelf:
```
OzzieIsaacs, line 522: here the query is right without _and

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
     return metadata
```
OzzieIsaacs, line 395: for the early PR: line 382 onwards:
```
        if sys.version_info < (3, 0):
            name = get_series(book).encode("utf-8")
        else:
            name = get_series(book)
        metadata["Series"] = {
            "Name": get_series(book),
```
looks suspicious to me to insert unencoded series and only use the decoded text to generate the id, could you please doublecheck it (I didn't do it up to now)

OzzieIsaacs: also following line:
```
       metadata["Series"] = {
            "Name": get_series(book),
            "Number": book.series_index,        # ToDo Check int() ?
            "NumberFloat": float(book.series_index),
```
From earlier commit. Could you please check: The book.series_index is a "float" like string with up to 2 decimals (meanwhile) shouldn't the books.series_index be casted to  int() ?

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+        "Items": [],
+        "LastModified": convert_to_kobo_timestamp_string(shelf.last_modified),
+        "Name": shelf.name,
+        "Type": "UserTag"
+    }
+    for book_shelf in shelf.books:
+        book = db.session.query(db.Books).filter(db.Books.id == book_shelf.book_id).one_or_none()
+        if not book:
+            log.info(u"Book (id: %s) in BookShelf (id: %s) not found in book database",  book_shelf.book_id, shelf.id)
+            return None
```
OzzieIsaacs, line 603: This code returns on the first not found book in a shelf, shouldn't it be `continue` instead, to get other books to be added in a shelf?

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+        name = shelf_request["Name"]
+        items = shelf_request["Items"]
+    except KeyError:
+        log.debug("Received malformed v1/library/tags request.")
+        abort(400, description="Malformed tags POST request. Data is missing 'Name' or 'Items' field")
+
+    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.name) == name, ub.Shelf.user_id ==
+                                              current_user.id).one_or_none()
+    if shelf and not shelf_lib.check_shelf_edit_permissions(shelf):
+        abort(401, description="User is unauthaurized to edit shelf.")
```
OzzieIsaacs, line 413: log message would be good, otherwise user has no idea why it's not working

OzzieIsaacs: invalid -> log is generated in shelf.py

cps/shelf.py
```diff
@@ -238,7 +240,7 @@ def create_shelf():
 @login_required
 def edit_shelf(shelf_id):
     shelf = ub.session.query(ub.Shelf).filter(ub.Shelf.id == shelf_id).first()
-    if request.method == "POST":
+    if request.method == "POST":            
         to_save = request.form.to_dict()
```
OzzieIsaacs, line 244: please remove trailing whitespaces

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+    shelf_request = request.json
+    name, items = None, None
+    try:
+        name = shelf_request["Name"]
+        items = shelf_request["Items"]
+    except KeyError:
+        log.debug("Received malformed v1/library/tags request.")
+        abort(400, description="Malformed tags POST request. Data is missing 'Name' or 'Items' field")
+
+    shelf = ub.session.query(ub.Shelf).filter(and_(ub.Shelf.name) == name, ub.Shelf.user_id ==
```
OzzieIsaacs, line 410: Big little problem, the name is not/no longer unique per user:
grafik
Every user can generate a public and private shelf with the same name, so we should check it during loading shelf name. I think we should prefer the private shelf over the public one with the same name. ( We are having 2 cases to consider 
1) name is not valid or not belonging to user
2) name is unique private or unique public shelf owned by user
3) name is used for private and public shelf of user
This problem occurs for ever shelf request with filtering from name

OzzieIsaacs:
I'm currently working on the PR (as you might have seen):

- [ ] Question regarding the missing feature of multiple authors: If there would be more than one author the filed would return a list of dicts similar to the DownloadUrls list, right?

- [ ]  Regarding "changing the pubdate also to a timestamp". From my point of view, it would be okay, I could modifiy the other parts in the code to handle the condequences, or is there a big issue you already know about?

OzzieIsaacs:
Start with fresh installation (means app.db is deleted before calibre-web starts and is freshly build). 
Activate upload feature in admin -> basic configuration -> feature configuration -> enable upload
Upload one book (in my case a epub file)
Edit the recently uploaded book
Push on delete button, you get this traceback (the line numbers may differ, as I did some code cosmetics):
```
 return func(*args, **kwargs)
File "/home/user/Entwicklung/calibre-web/cps/editbooks.py", line 182, in delete_book
ub.session.query(ub.ArchivedBook).filter(ub.ReadBook.book_id == book_id).delete()
File "/home/user/.local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 3689, in delete
delete_op.exec_()
File "/home/user/.local/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 1672, in exec_
self._do_pre_synchronize()
File "/home/user/.local/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 1751, in _do_pre_synchronize
"synchronize_session parameter." % err
InvalidRequestError: Could not evaluate current criteria in Python: "Can't evaluate criteria against alternate class <class 'cps.ub.ReadBook'>". Specify 'fetch' or False for the synchronize_session parameter.
```
THe archived books table looks like this:
grafik

cps/kobo.py
OzzieIsaacs: Another idea would be: In the UI to the public shelfs the text 'public' is appended, so maybe doing this here would also be possible. But this text is translatable, so it depends on the user interface language.

cps/kobo.py
```diff
@@ -392,6 +394,222 @@ def get_metadata(book):
...
+    return make_response('', 200)
+
+
+# Add new, changed, or deleted shelves to the sync_results.
+# Note: Public shelves that aren't owned by the user aren't supported.
+def sync_shelves(sync_token, sync_results):
+    new_tags_last_modified = sync_token.tags_last_modified
+
+    for shelf in ub.session.query(ub.ShelfArchive).filter(func.datetime(ub.ShelfArchive.last_modified) > sync_token.tags_last_modified, ub.ShelfArchive.user_id == current_user.id):
+        new_tags_last_modified = max(shelf.last_modified, new_tags_last_modified)
```
OzzieIsaacs, line 556: Haven't fully understood what we are doing there.
I did the following: Generate shelf on UI (private one), sync it to kobo (New Tag is synced), add book to shelf, try to sync it again.
Expected result: Changed tag is synced, but I get empty response.
Upon looking into ShelfArchive table I find this:
grafik
User ID is identical to the uuid  but it's compared to the internal Calibre-Web user iD, which is a number from 1 to x, so comparsion fails, and the book in shelf isn't synced.

shavitmichael:
Wow lot's of great comments, thanks for taking a look!

> * [ ]  Question regarding the missing feature of multiple authors: If there would be more than one author the filed would return a list of dicts similar to the DownloadUrls list, right?

Here's an example from the official Kobo store:
```
"BookMetadata" : {
  "ContributorRoles": [
    {
      "Name": "Garth Nix"
    },
    {
      "Name": "Leo and Diane Dillon"
    }],
  "Contributors": [ "Garth Nix", "Leo and Diane Dillon"],
```
Both authors show up under the Book tittle on my device, and both authors have an entry on my device's **Author** tab.

> * [ ]   Regarding "changing the pubdate also to a timestamp". From my point of view, it would be okay, I could modifiy the other parts in the code to handle the condequences, or is there a big issue you already know about?

I don't think there should be any issues, the usages are pretty spread out throughout the code-base so it's some work to update everything.

cps/kobo.py
shavitmichael: I'm removing the and_ clause from all these filter calls in 06c15a7 .
The reason this works is because the filter function joins together it's argument into an and_ clause by default.

cps/kobo.py
shavitmichael: Fixed in 06c15a7

cps/kobo.py
shavitmichael: Fixed in 06c15a7

cps/kobo.py
shavitmichael: Fixed in 06c15a7

cps/kobo.py
shavitmichael: Fixed in 06c15a7

cps/kobo.py
shavitmichael: Fixed in 06c15a7

cps/kobo.py
shavitmichael: Indeed, the check_shelf_edit_permissions call outputs an error to log.error

cps/shelf.py
shavitmichael: Fixed in 06c15a7

cps/kobo.py
shavitmichael: This particular section is propagating deletions of shelves from CalibreWeb to the device. And indeed it looks like there's a bug in the delete_shelf_helper() I added to shelf.py; specifically line 286
```
    ub.session.add(ub.ShelfArchive(uuid = cur_shelf.uuid, user_id = cur_shelf.uuid))
```
With this bug, deleting a shelf from CalibreWeb and syncing the device doesn't remove the shelf form the device.
With 9296d35 , deleting a shelf from CalibreWeb will create an entry in the ShelfArchive table with the correct user_id field. Deleting a shelf form CalibreWeb also removes it from the device after syncing.

cps/kobo.py
shavitmichael: I might be misunderstanding, but I don't think this is an issue for us.

We only use the shelf name during shelf creation. Other API calls (such as adding/removing books from a shelf, or deleting a shelf) rely on the ub.Shelf.uuid field to identify the shelf they operate on.
The only reason we check the shelf name here is to prevent the user from **creating** multiple shelves with the same name. However if shelves with the same name do exist, that won't affect the Kobo sync protocol in any way.

shavitmichael:
> Start with fresh installation (means app.db is deleted before calibre-web starts and is freshly build).
> Activate upload feature in admin -> basic configuration -> feature configuration -> enable upload
> Upload one book (in my case a epub file)
> Edit the recently uploaded book
> Push on delete button, you get this traceback (the line numbers may differ, as I did some code cosmetics):
> 
> ```
>  return func(*args, **kwargs)
> File "/home/user/Entwicklung/calibre-web/cps/editbooks.py", line 182, in delete_book
> ub.session.query(ub.ArchivedBook).filter(ub.ReadBook.book_id == book_id).delete()
> File "/home/user/.local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 3689, in delete
> delete_op.exec_()
> File "/home/user/.local/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 1672, in exec_
> self._do_pre_synchronize()
> File "/home/user/.local/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 1751, in _do_pre_synchronize
> "synchronize_session parameter." % err
> InvalidRequestError: Could not evaluate current criteria in Python: "Can't evaluate criteria against alternate class <class 'cps.ub.ReadBook'>". Specify 'fetch' or False for the synchronize_session parameter.
> ```
> 
> The archived books table looks like this:
> grafik

Good catch! Fixed in 742ec2b .

cps/opds.py
```diff
@@ -326,12 +326,12 @@ def feed_shelfindex():
 def feed_shelf(book_id):
     off = request.args.get("offset") or 0
     if current_user.is_anonymous:
-        shelf = ub.session.query(ub.Shelf).filter(ub.Shelf.is_public == 1, ub.Shelf.id == book_id).first()
+        shelf = ub.session.query(ub.Shelf).filter(ub.Shelf.is_public == 1, ub.Shelf.id == book_id, not ub.Shelf.deleted).first()
     else:
```
OzzieIsaacs, line 330: We almost got it:
grafik
But there is one sideeffect: This line and line 334 causes a crash upon access. Because there is no column deleted in Shelf. I currently can't oversee what the intention of the change was. I would delete it?

OzzieIsaacs:
During testing found some more or less big problems and fixed them by myself:
Creating a tag via Kobo interfaces failed, because uuid() doesn't return a string
Updating Shelfs from the UI didn't trigger a sync, because the last_modified time wasn't updated.
I improved error handling to catch empty post requests missing elements in every request and NaN's in the reading status
I removed the 404 errorhandler, because it hijacked the general 404 handler of the project and so every invalid request would have proxied to kobo, or would have been tranfered to a empty "Code 200 {}" request
Empty shelf names are prevented to create over the sync interface

PR #1267, opened 2020-03-22
Fix metadata display when there's results from comicvin

xcffl:
When there is some results from comicvin, the current logic will replace the book list with `nothing found`, since `showResults()` revoked previously has reset `ggData` or `dbData` to `0`.


OzzieIsaacs:
I don't think you code works if any of the other metadata provider returns no result

PR #1277, opened 2020-03-30
[Kobo] Fix issue with covers

shavitmichael:
- The image_url_template and image_url_quality_template do not resolve correctly for me when wsgi_app.is_proxied is true. I serve the library under https://<ip>/calibre, and the image_url_template would end up as <ip>/calibre/calibre/etc.
- We weren't generating a request to the correct endpoint when redirecting cover requests to the official Kobo store.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -535,16 +535,16 @@ def get_current_bookmark_response(current_bookmark):
     return resp
 
 
-@kobo.route("/<book_uuid>/image.jpg")
+@kobo.route("/<book_uuid>/<width>/<height>/image.jpg")
 @requires_kobo_auth
-def HandleCoverImageRequest(book_uuid):
+def HandleCoverImageRequest(book_uuid, width, height):
     book_cover = helper.get_book_cover_with_uuid(
         book_uuid, use_generic_cover_on_failure=False
     )
     if not book_cover:
         if config.config_kobo_proxy:
             log.debug("Cover for unknown book: %s proxied to kobo" % book_uuid)
-            return redirect(get_store_url_for_current_request(), 307)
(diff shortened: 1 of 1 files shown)


cps/kobo.py
```diff
@@ -678,33 +683,42 @@ def HandleInitRequest():
             url_base=host,
             url_port=config.config_port
         )
+        kobo_resources["image_host"] = calibre_web_url
+        kobo_resources["image_url_quality_template"] = unquote(calibre_web_url + url_for("kobo.HandleCoverImageRequest",
+                                                                                         auth_token=kobo_auth.get_auth_token(),
+                                                                                         book_uuid="{ImageId}",
+                                                                                         width="{width}",
+                                                                                         height="{height}",
+                                                                                         _external=True))
```
OzzieIsaacs, line 692: The line external=TRUE gave me wrong urls the hostname was doubled, I removed it

cps/kobo.py
```diff
@@ -678,33 +683,42 @@ def HandleInitRequest():
...
+        kobo_resources["image_url_template"] = unquote(calibre_web_url + url_for("kobo.HandleCoverImageRequest",
+                                                                                 auth_token=kobo_auth.get_auth_token(),
+                                                                                 book_uuid="{ImageId}",
+                                                                                 width="{width}",
+                                                                                 height="{height}",
+                                                                                 _external=True))
```
OzzieIsaacs, line 698: The line external=TRUE gave me wrong urls, the hostname was doubled, I removed it

cps/kobo.py
```diff
@@ -535,16 +535,16 @@ def get_current_bookmark_response(current_bookmark):
     return resp
 
 
-@kobo.route("/<book_uuid>/image.jpg")
+@kobo.route("/<book_uuid>/<width>/<height>/image.jpg")
 @requires_kobo_auth
```
OzzieIsaacs, line 539: I added the unused "Quality" and "isGreyscale" paths to the route, so we are nearer to Kobo's image url and this causes less trouble in the future

PR #1282, opened 2020-03-31
Updated to exclude new modules

hexeth:
Thanks Becky


OzzieIsaacs:
Thanks a lot for the patch

PR #1287, opened 2020-04-02
Attempt to allow using other dbs than sqlite for settings

halkeye:
Small patch that would allow settings to be stored in something other than sqlite.

My feeling is if you don't document it, then you don't have to support it, but it'll allow people to do it.

My Use case:
I run most things in a kubernetes cloud. I have postgres backed up regularly. I want to run calibre-web, but the app needs the tiny sqlite file persisted, which means I have to pay for disk. Dropping the need for the sqlite file, means it can run diskless (google drive, or other mount for books).

Tested via docker

Dockerfile
```
FROM linuxserver/calibre-web
RUN apt-get update && apt-get install -y patch python3-pip libpq-dev
RUN pip3 install psycopg2
ADD patch.patch .
RUN patch -p1 -d /app/calibre-web < patch.patch
```

docker-compose.yaml
```
version: "3.7"

services:
  app:
    build: .
    environment:
      - SETTINGS_DB_ENGINE_URI=postgres://admin:password@database:5432/calibre-web
    links:
      - database
    depends_on:
      - database
    ports:
      - "3000:8083"
    volumes:
      - ./data:/books:ro

  database:
    image: postgres:11.4-alpine
    environment:
      - POSTGRES_USER=admin
      - POSTGRES_PASSWORD=password
      - POSTGRES_DB=calibre-web
```

```
git diff > patch.patch; docker-compose up --build
```

### DIFF
--- a/cps/ub.py
+++ b/cps/ub.py
@@ -31,7 +31,7 @@ try:
     oauth_support = True
 except ImportError:
     oauth_support = False
-from sqlalchemy import create_engine, exc, exists
+from sqlalchemy import create_engine, exc, exists, inspect
 from sqlalchemy import Column, ForeignKey
 from sqlalchemy import String, Integer, SmallInteger, Boolean, DateTime
 from sqlalchemy.orm import relationship, sessionmaker
@@ -296,7 +296,7 @@ class Bookmark(Base):
     id = Column(Integer, primary_key=True)
     user_id = Column(Integer, ForeignKey('user.id'))
     book_id = Column(Integer)
-    format = Column(String(collation='NOCASE'))
+    format = Column(String())
     bookmark_key = Column(String)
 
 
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
I checked it with normal python. There is a database initializing problem if you use the normal storage. If you delete the app.db file, start the configuration(with normal sqlite storage) you can add the library path, but if you try to login in afterwards you get wrong username/passwort. If you stop the program and start again it works, password and username are accepted.

halkeye:
okay i'll poke at it and see what I can fix

OzzieIsaacs:
> My feeling is if you don't document it, then you don't have to support it, but it'll allow people to do it.

Meanwhile I think it's not a good idea to do it, there are many people doing crazy things with the software, having issues, complain about it and don't see what they are doing what isn't standard, and don't document this in there issues. So I think, even if you get it running, I will not merge it.
So I decided to close this PR

Issue #1307, opened 2020-04-14
github oauth login not visible


flashlab:
**Describe the bug/problem**
I use the nightly version, and have installed optional-requirements related oauth, but in the setup page I can't find the input option.
Also int the login page, there is no login with github button.

**Environment (please complete the following information):**
 - OS: Debian 9
 - Python version python 3.5.3
 - Calibre-Web version latest
 - Browser chrome

OzzieIsaacs:
Looks like 
flask-dance>=0.13.0
was to optimistic.
If you install version 0.13 it works. I've to check what the problem is.

OzzieIsaacs:
Should be visible back with the newest commit

flashlab:
@OzzieIsaacs Thanks for your patch, the github login button come back, and I can setup the app id and secret.
Now the problem is callback url.
I notice the redirect_uri field `http://127.0.0.1:8083/login/github/authorized` I can only set the same value in github's `Authorization callback URL` field before correctly jump to github authorization page.
but I host calibre-web with domain, not locally, so how can I make the redirect_uri use the real domain? Thak you very much.

OzzieIsaacs:
You need a public accessible instance of Calibre-Web. So if your computer is accessible from internet via "example.org:8083", please access Calibre-Web via this address, afterwards the callback address will magically change to `http://example.org:8083/login/github/authorized`. This is how it should work. Once authorized and logged in you can access your Calibre-Web instance again via 127.0.0.1...

flashlab:
@OzzieIsaacs Sorry for not making my situation clear. I host Calibre-Web on VPS and use Nginx to proxy 8083 port to my domain https://calibre.redbux.cn 
For now, I can use `ip:port format` (like this: `http://xxx.xxx.xxx.xxx:8083/login/github/authorized`) in the `Authorization callback URL` field to jump to github authorization page, but after click the button and redirect back, I receive `500 Internal Server Error`:
```
Traceback (most recent call last):
File "/home/zjf/calibre-web/vendor/flask/app.py", line 2447, in wsgi_app
response = self.full_dispatch_request()
File "/home/zjf/calibre-web/vendor/flask/app.py", line 1952, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/home/zjf/calibre-web/vendor/flask/app.py", line 1821, in handle_user_exception
reraise(exc_type, exc_value, tb)
[...]

PR #1318, opened 2020-04-19
Fix https github oauth

flashlab:
@OzzieIsaacs After spending several hours digging the github oauth problem #1307 , I finally makes it work with https domains.
First, the nginx configuration is needed to properly proxy the real domain and protocol to Flask, following [this example](https://flask.palletsprojects.com/en/1.0.x/deploying/wsgi-standalone/#deploying-proxy-setups).
And pyopenssl model is required: `pip3 install -U pyopenssl`.
Finally, import `ProxyFix` to make flask properly identity https protocols.
Feel free to test here https://calibre.redbux.cn
Some useful links：
https://stackoverflow.com/questions/54278247/flask-dance-and-error-redirect-uri-mismatch
https://flask-dance.readthedocs.io/en/latest/proxies.html
https://github.com/singingwolfboy/flask-dance/issues/188

### DIFF
--- a/cps/__init__.py
+++ b/cps/__init__.py
@@ -36,7 +36,7 @@ from flask_principal import Principal
 from . import logger, cache_buster, cli, config_sql, ub, db, services
 from .reverseproxy import ReverseProxied
 from .server import WebServer
-
+from werkzeug.middleware.proxy_fix import ProxyFix
 
 mimetypes.init()
 mimetypes.add_type('application/xhtml+xml', '.xhtml')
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
I had time to doublecheck the PR. 
By just using the example given in the Calibre-Web wiki, replacing the `/calibre` with `/` and adding ssl to the nginx configuration it worked out of the box. Because werkzeug is talking about security concerns in using the proxyfix function and it shall only be used if it is necessary, I reverted your PR. I didn't have to install anything additional. I've tested it with an up to date Manjaro Linux with nginx 1.17 ("mainline"), the ssl Certificates where signed by Let's encrypt

PR #1337, opened 2020-04-29
Don't use an hardcoded session key

jvoisin:
This fixes a trivial authentication bypass,
according to https://flask.palletsprojects.com/en/1.1.x/quickstart/#sessions

### DIFF
--- a/cps/__init__.py
+++ b/cps/__init__.py
@@ -89,7 +89,7 @@ def create_app():
     log.info('Starting Calibre Web...')
     Principal(app)
     lm.init_app(app)
-    app.secret_key = os.getenv('SECRET_KEY', 'A0Zr98j/3yX R~XHH!jmN]LWX/,?RT')
+    app.secret_key = os.getenv('SECRET_KEY', os.urandom(32))
 
     web_server.init_app(app, config)
     db.setup_db(config)


OzzieIsaacs:
I agree with you, the hardcoded session key isn't good. Your solution leads to an automatic logout after restart of calibre-web. 
This is okay so far, except one usecase: If you configure Calibre-Web in admin section there are several configurations, which lead to a reboot of Calibre-Web to apply the new setting (admin.py, functions update_view_configuration and _configuration_update_helper). In this cases you have to login after applying each setting. 
I suggest storing the session_key in the app.db on first start of Calibre-Web and using the same session key after each restart. Or to be more secure: Only in cases of an intentional restart the session key is stored in the database, the boot logic loads the session key afterwards, deletes it in the app database. For any other reboots the key isn't stored and after restart of Calibre-Web a new session key is generated.

jvoisin:
This is tracked as [CVE-2020-12627]( https://nvd.nist.gov/vuln/detail/CVE-2020-12627 ).

OzzieIsaacs:
@technosoft2000: There was security problem found in Calibre-Web. I created a release with version 0.6.7 to address the issue. The problem could be avoided in older releases by generating an own session key and providing it to Calibre-Web via environment variable 'SECRET_KEY'

Technosoft2000:
Thank you for the information @OzzieIsaacs

PR #1368, opened 2020-05-09
Improve a bit the cookie's hardening

jvoisin:
- Samesite for session cookies as well as the remember me ones
- Httponly


jvoisin:
I used `SameSite=Lax` instead of `Samesite=Strict`, because I wasn't sure if it was going to break some stuff.

Issue #1357, opened 2020-05-06
Kepub books not renamed to kepub.epub when downloading


gpulido:
When I try to download an already converted kepub book on calibre onto my Kobo device using the experimental browser the kobo doesn't download it as the kepub's are stored in calibre as .kepub and kobo needs them to be .kepub.epub (it is just a rename of the file adding the epub)
Several solutions:
1) add a flag to calibre web to "download kepub as kepub.epub" so all downloads are renamed. 
2) whenever a kepub format is found add another option to the download button to download the kepub.epub

I known there is some work done / being done onto the kobo sync functionality but this is "unrelated" as is just a way to download the kepub with the right extension.
Please forgive me if this has already asked or if there is any kind of workaround to solve this. I haven't found it yet.

OzzieIsaacs:
I understood the problem, and I agree with you, this is not related to the kobo sync stuff

gpulido:
Hello,
I'm reviewing the code and I will try to make a pr for this.
It seems very easy just to attach the .epub to the filename tag on the response headers if the fileformat is a Kepub. But that would mean that any download from the web of a kepub format it will include the kepub.epub.
At the end kepub is only used by kobo and it needs such extension, so maybe it make sense (is the same that is doing calibre when you send a kepub file to a kobo device), but it seems a bit patch to just do something like this (look to the real_book_format)
```python

def get_download_link(book_id, book_format):
    book_format = book_format.split(".")[0]
    book = db.session.query(db.Books).filter(db.Books.id == book_id).filter(common_filters()).first()
    if book:
        data1 = db.session.query(db.Data).filter(db.Data.book == book.id)\
            .filter(db.Data.format == book_format.upper()).first()
    else:
        abort(404)
    if data1:
        # collect downloaded books only for registered user and not for anonymous user
        if current_user.is_authenticated:
            ub.update_download(book_id, int(current_user.id))
        file_name = book.title
        if len(book.authors) > 0:
            file_name = book.authors[0].name + '_' + file_name
        file_name = get_valid_filename(file_name)
        real_book_format = book_format + '.epub' if book_format == 'kepub' else book_format
        headers = Headers()
        headers["Content-Type"] = mimetypes.types_map.get('.' + book_format, "application/octet-stream")
        headers["Content-Disposition"] = "attachment; filename=%s.%s; filename*=UTF-8''%s.%s" % (
[...]

PR #1375, opened 2020-05-10
Append .epub extension when downloading .kepub format

gpulido:
This will make that  when an Kepub file is being asked to download an .epub extension is added automatically.
## Motivation
Calibre stores kepub files as .kepub to not messing them with the .epub format.
When calibre sends a kepub file to a kobo device it automatically appends the .epub extension so the name becomes "name_of_the_book.kepub.epub" that is the way kobo devices need to be named.
When a kobo device is using the experimental browser to access calibre-web and try to download a kepub it doesn't have the oportunity to rename the file to use kepub.epub. This pr solve that.
If any other kind of device is trying to download (for example using a desktop pc and a browser) the browser download feature will show a dialog where the user could be able to change the name of the file (and remove the epub part if the user wants). 
fixes https://github.com/janeczku/calibre-web/issues/1357

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -890,10 +890,11 @@ def get_download_link(book_id, book_format):
         if len(book.authors) > 0:
             file_name = book.authors[0].name + '_' + file_name
         file_name = get_valid_filename(file_name)
+        real_book_format = book_format + '.epub' if book_format == 'kepub' else book_format
         headers = Headers()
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
I'm having a solution from another pull request based on a client detection algorithm. It's merged to the development branch.

Thank you anyhow for the contribution

PR #1425, opened 2020-05-24
Update messages.po

rafarq:
For many words there is a big difference with Spanish from Spain, and American Spanish. This file is in Spanish from Spain now. I will continue adding strings later but you might consider adding Spanish flavors to the interface.


rafarq:
This is my first pull request, I hope i did it well

PR #1429, opened 2020-05-25
Update messages.po

rafarq:
Here is the whole file traslated


rafarq:
Hope I did it well, it my first time doing a pull request.

OzzieIsaacs:
Thanks for the contribution. 
Regarding the different country translations: I need the help of the translator to tell my which "dialect" of a language is used, I can then look for the right translation version. Valid versions for Spanish would be:
grafik
But I'd like to point out, that translating is an ongoing process, so to many different "dialects" need much more effort, than just one language version

OzzieIsaacs:
Regarding doing it right: Almost. 
Please do not provide any translations to the development branch, as this causes trouble. I merged this one to the master branch, therefore I',m now closing the PR and it's not marked as merged (even if it is)

PR #1433, opened 2020-05-26
Add some static analysis badges

jvoisin:


OzzieIsaacs:
Thanks for the suggestions, but as I mentioned earlier, I'm using codacy.

jvoisin:
LGTM and coverity and finding some issues that codacy doesn't find.

PR #1443, opened 2020-05-28
Add series name and index to book display

Knepherbird:
First big pull request! Forgive me if I mess something up ;) 

While thinking about making book display info more meaningful for my comic book collection, I decided to start small and add Series info to the book display for any format. As a default for regular books, I decided to make it subtle. I'm undecided between a simple Series Name ([index]) vs. Book [index] of Series Name. Also, converting the series index to an int is cleaner, but there are some series with x.5 entries. 

calibre-web-show-series-meta-v2


b-t-k:
Great addition. I've been doing this manually with every update.

Knepherbird:
One thing that would be nice would be to suppress the ".0" and only display integer values unless the series index is something like .1 or .5. These are rare for book series, but some of the novellas and side stories in The Expanse are numbered this way, and there are some comics that have had "point" issues. This is the first time I've worked on anything built on python, so I'm not sure the best way to go about dropping in a function that would check and format the series index.

RottenRonnie:
Would it be possible to add this to the "Authors" view as well? 
I was hoping to paste in the following into authors.html but have no idea where it can be safely inserted to show series and number in the author view. 

cps/templates/author.html


        {% if entry.series.__len__() > 0 %}
        <p class="series">
          <a href="{{url_for('web.books_list', data='series', sort='new', book_id=entry.series[0].id )}}">
            {{entry.series[0].name}}
          </a> 
          ({{entry.series_index}})
        </p>
        {% endif %}

Knepherbird:
Looks like that will require updating the database query for the "author" web route call in web.py. When I have some free time, I'll look into it. I'm primarily a front-end dev, and have never worked with python before I started poking around in this project, so I've got some catching up to do.

DruidGreeneyes:
Bump? I went ahead and hacked out a version of this locally, but it would be nice to be able to get it from upstream instead.

PR #1458, opened 2020-05-31
Some czech translations

flying-sausages:
I could not find any guide on contributing translations but here are some strings. Not sure if someone needs to be compiled or not.


OzzieIsaacs:
Thanks

PR #1459, opened 2020-06-01
[enhancement] cleaned up the setup.py install script

epsilon-0:
Adding the work that I've done while porting to OpenBSD, its working nicely at https://lib.epsilonknot.xyz

This allows a default installation target and makes the setup.py script work again perfectly.

Caveats (most of them unrelated to this PR):
1)  the installation is not a module anymore, and can not be run as a simple
```
python3.7 -m cps.py
```
Instead I have to run it as 
```
python3.7 /usr/local/lib/python3.7/site-packages/cps.py
```
Unfortunately I don't have the skillset for getting this working :crying_cat_face: 

2) The installation cannot be run as a regular user as during startup calibre-web tries to create files in the installation directory.  
Current workaround - run as root :no_entry_sign:  
Desired fix - don't check for permissions in the install directory, but ask for where to read/write things

Again, I'm not sure how to get this working :crying_cat_face:

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,1 @@
-graft src/calibreweb
+graft cps
--- a/setup.py
+++ b/setup.py
@@ -21,26 +21,20 @@
 
 from setuptools import setup
 from setuptools import find_packages
-import os
 import re
-import codecs
+import ast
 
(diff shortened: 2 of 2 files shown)


epsilon-0:
BTW I know almost nothing about python so feel free to modify the PR to make it better.

PR #1467, opened 2020-06-04
Update generate_kobo_auth_url.html

Dinth:
i believe this was a typo


OzzieIsaacs:
Thanks for the PR. I think it's indeed inconsistent, I removed the quotes totally, so it's easier to copy the string

PR #1480, opened 2020-06-09
Updated CZ translation

nykul:


OzzieIsaacs:
Thanks for the translation.
One remark: If you create a pull request, please leave the files where they are belong and update them in place, otherwise the merge is getting complicated and confusing, resulting in mistakes.

PR #1487, opened 2020-06-11
Fix up Templates for issues

flying-sausages:


OzzieIsaacs:
Thanks a lot, was indeed a bit strange

PR #1516, opened 2020-07-03
feat(api): include external port option

jef:
### Description

This should help with a fix when downloading from a URL that goes through a reverse proxy, especially through Docker.

As there could be port mappings within Docker (`12345:8083`), then a reverse proxy: (`54321 -> 12345`), and even potential router port forwarding (`34125 -> 54321`). This allows the user to make API calls correctly to links like `api_endpoint` for Kobo.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -623,6 +623,7 @@ def _configuration_update_helper():
         gdriveError = _configuration_gdrive_helper(to_save)
 
         reboot_required |= _config_int(to_save, "config_port")
+        reboot_required |= _config_int(to_save, "config_external_port")
 
         reboot_required |= _config_string(to_save, "config_keyfile")
         if config.config_keyfile and not os.path.isfile(config.config_keyfile):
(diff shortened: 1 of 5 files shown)


jef:
If we don't think this could go under the main Server Configuration (since it seems to be mostly related to Kobo), we can put this option under **Feature Configuration** -> **Enable Kobo sync** as well. But I figured that this could help other features in the future.

OzzieIsaacs:
I had time to double check it, there is indeed not chance for detecting the port. One thing I don't conclude with you is when it is needed. In my opinion it is needed in case of port forwarding, and (now) in case of direct connection with non standard ports. For reverse proxy configurations, the code is detecting the reverse proxy and the malformed kobo request isn't getting to calibre-web.
But anyhow, I did some small changes to your code:
- I moved the setting near to kobo setting, as it's only relevant in case kobo sync is used
- I removed the reboot necessary in the save section
- I changed the text according to my conclusion when it is needed.

Thanks for the work

jef:
You bet! Thanks for the update and merge. You have much more insight to how calibre-web works, so I really appreciate the help! I'll pull down the latest this evening and hopefully can close out a few issues.

Issue #1534, opened 2020-07-22
Allow to define user nickname with a LDAP field in user import 


cpoissonnier:
**Is your feature request related to a problem? Please describe.**

My problem is pretty much the same as this [issue](https://github.com/janeczku/calibre-web/issues/1389#issuecomment-627346691).

When I request LDAP to get the members of my group, I have a list of user that matches this format :
`CN=Firstname LastName,OU=...OU=...,DC=...,DC=...`

Unfortunatly, `sAMAccountName` field is not present, so I have to use `(&(objectclass=Person)(cn=%s))` as  LDAP User Object Filter, and user nickname is set with CN, which is not the login I want to use (cf [this instruction](https://github.com/janeczku/calibre-web/blob/master/cps/web.py#L350)).

**Describe the solution you'd like**
It would be great to allow us to define what field we want to use as user nickname in LDAP user import.

We could say : "ok, you have to use `(&(objectclass=Person)(cn=%s))` to fetch the user, but the login will be the value in `sAMAccountName` field.

cpoissonnier:
#1539

PR #1539, opened 2020-07-23
feat(ldap): add a field that allows to override LDAP User Object Filter when a user is imported

cpoissonnier:
I created [this feature request yesterday](https://github.com/janeczku/calibre-web/issues/1534), and since I had some free time to implement it, there you go :)

The goal of this pull request is to allow admins to use another LDAP filter to import users, and to use a field as nickname that is not necessarily in the member field of the group.

In my case, I had this format in member field : 
`CN=Firstname LastName,OU=...OU=...,DC=...,DC=...`
but I wanted to use sAMAccountName as login in Calibre-Web.

Note 1 : this is the first time EVER I code in Python, so I might have missed some points
Note 2 : I will be unavailable from 25 july to 10 august

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -564,6 +564,7 @@ def _configuration_ldap_helper(to_save, gdriveError):
     reboot_required |= _config_string(to_save, "config_ldap_user_object")
     reboot_required |= _config_string(to_save, "config_ldap_group_object_filter")
     reboot_required |= _config_string(to_save, "config_ldap_group_members_field")
+    reboot_required |= _config_string(to_save, "config_ldap_member_user_object")
     reboot_required |= _config_checkbox(to_save, "config_ldap_openldap")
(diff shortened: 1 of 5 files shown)


OzzieIsaacs:
I had time to look at your PR and merged it. I modified it a bit: 
- I set the config value for config_ldap_member_user_object to empty string, so the default behavior is "autodetect". 
- I added a select to choose between "autodetect" and "enter configuration" for config_ldap_member_user_object
- I removed the "app.config['LDAP_MEMBER_USER_OBJECT_FILTER'] = config.config_ldap_member_user_object" line in simpleldap, as there is no such config option in flask-simpleldap
- Reverted your change in line 338-345 in web.py. If we have a vaild username and compare it with all users in database, there is no reason for doing more user handling than in the login routine. This part I'm only 99.9% sure. If I'm wrong, please correct me (we then have to revert my reverted changes).

grafik

Issue #1531, opened 2020-07-19
PDF cover generation makes transparent backgrounds black


shibacomputer:
**Describe the bug/problem**
When uploading a PDF, the Imagemagick's automatically generated thumbnail interperets background transparency as black. For many PDF documents, this makes them illegible when browsing a collection.

**To Reproduce**
1. Click upload
2. Choose a PDF with a large amount of transparency on the first page
3. Confirm upload
4. Once the metadata screen has loaded, the black background jpg thumbnail is visible.

**Logfile**
Not sure I need this, given that it's not erroring out?

**Expected behavior**
Background transparency should assume white, instead of black.

**Screenshots**
![image](https://user-images.githubusercontent.com/737950/87871911-bb2ff500-c9b4-11ea-8103-2b982c41224f.png)

**Environment (please complete the following information):**
 - OS: Ubuntu Server 20.04 LTS
 - Python version: python 3.8.2
 - Calibre-Web version: 0.6.8
 - Docker container: Linuxuser
 - Special Hardware: n/a
 - Browser: Tested on Firefox 78.0.2 Windows 10 & macOS Catalina

PR #1540, opened 2020-07-23
Fix background black assumption

shibacomputer:
This is a 3 line fix for #1531 

1. Import `Color` from `wand.image` (alongside `Image`)
2. In `pdf_preview`, add two additional properties to remove the alpha channel (since we're exporting as jpg), and declare `img.background_color` as white.

### DIFF
--- a/cps/uploader.py
+++ b/cps/uploader.py
@@ -36,6 +36,7 @@ except ImportError:
 
 try:
     from wand.image import Image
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Sorry, you've been a bit too late. I programmed it yesterday and  just forgot to commit my changes. 
Thanks you

PR #1562, opened 2020-08-09
Trim whitespace from filename when uploading

blitzmann:
I was having an issue uploading a file, getting a permission denied error. It wasn't a permissions thing, instead the metadata of the book apparently has whitespace at the end of the author. So when concating the folder structure, it looks something like this: `calibrelib\Author Name \Book Title.epub`

Note the whitespace at the end of the author. This was causing the error. I simply strip whitespace from the ends.

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -254,7 +254,7 @@ def get_valid_filename(value, replace_whitespace=True):
         value = re.sub(r'[\*\+:\\\"/<>\?]+', u'_', value, flags=re.U)
         # pipe has to be replaced with comma
         value = re.sub(r'[\|]+', u',', value, flags=re.U)
-    value = value[:128]
+    value = value[:128].strip()
     if not value:
         raise ValueError("Filename cannot be empty")
     if sys.version_info.major == 3:


OzzieIsaacs:
The patch you provide is not the right one for this problem:
The spaces are striped in line 242 or 249/251 if you are not using Unidecode
Is it possible that the authors name was longer than 128 characters and was truncated and the space was caused by this?

blitzmann:
Ah, indeed!

The raw data that is fed into that function is:
`Marilyn Allman Maye| Harold S. Buchanan| Jannette O. Domingo| Joyce Frisby Baynes| Marilyn Holifield| Myra E. Rose| Bridget Van Gronigen Warren ` 

Right before truncation:
`'Marilyn Allman Maye, Harold S. Buchanan, Jannette O. Domingo, Joyce Frisby Baynes, Marilyn Holifield, Myra E. Rose, Bridget Van Gronigen Warren'

So indeed, it looks like the truncation is truncating `Gronigen Warren`, but then leaving the space a the end of the result. 

Regardless, even tho it's being stripped before this part, it should be done as the very last step to catch weird cases like this. Or should I remove the other `.strip()`s and just rely on the one added with this commit? I think the other `.strip()`s were to catch things like this, but then maybe truncation was added later and caused this bug?

Let me know how you would like me to proceed

OzzieIsaacs:
I think our real problem is that it's not detecting we are dealing with 7 different authors. Let me think a bit about this, because I think it's not so easy at it looks on the first view:
Maye, Marilyn Allman & Buchanan, Harold S.
or
Maye, Marilyn Allman ; Buchanan, Harold S.

would be 2 Authors and not 4

blitzmann:
For what it's worth, when importing it via Calibre directly, the authors are:
```
Marilyn Allman Maye, Harold S. Buchanan, Jannette O. Domingo, Joyce Frisby Baynes, Marilyn Holifield, Myra E. Rose, Bridget Van Gronigen Warren & Aundrea White Kelley.
```

and the path that Calibre creates is `lib\Marilyn Allman Maye, Harold S. Bucha\Seven Sisters and a Brother (4)`

I'm not sure I'm following what you're trying to describe tho. What would the number of authors have anything to do with it or this issue? Just looking for more clarification there :)

OzzieIsaacs:
If Calibre-Web would detect the authors correct, each author‘s name length would be way less than 128 characters, so it wouldn‘t be truncated. 
What I wanted to say with my example is that just splitting it at the ',' wouldn't work

blitzmann:
I will check and confirm later tonight, thanks!

Issue #1564, opened 2020-08-11
Random errors about tables not existing - narrowed down to CalibreDB thread


blitzmann:
This is both a bug that I've seen when converting books, but also kind of a development support question as I'm hacking around in the source code to get a better feel for it so that I can perform some customization's.

The bug in question is that I get random errors about `books` or `authors` table not existing when converting formats. This is infrequent, and somewhat hard to reproduce with any sort of consistency unfortunately. There also seems to be two places where it might break: 
1) In the worker thread when calibre-web gets the book (most common) - seems to maybe happen when you try to convert multiple things at a time? 
2) The CalibreDB thread where it tries to add the link to the new format (less common).

Exception from the worker thread:
```
Exception in thread Thread-3:
Traceback (most recent call last):
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\venv\lib\site-packages\sqlalchemy\engine\base.py", line 1278, in _execute_context
    cursor, statement, parameters, context
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\venv\lib\site-packages\sqlalchemy\engine\default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: no such table: books

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\cps\worker.py", line 216, in run
    self._convert_any_format()
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\cps\worker.py", line 272, in _convert_any_format
    filename = self._convert_ebook_format()
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\cps\worker.py", line 298, in _convert_ebook_format
    cur_book = calibre_db.get_book(book_id)
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\cps\db.py", line 475, in get_book
    return self.session.query(Books).filter(Books.id == book_id).first()
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\venv\lib\site-packages\sqlalchemy\orm\query.py", line 3397, in first
    ret = list(self[0:1])
  File "C:\Users\holme\Documents\Sync\Git\blitzmann\calibre-web2\venv\lib\site-packages\sqlalchemy\orm\query.py", line 3171, in __getitem__
[...]

PR #1565, opened 2020-08-11
WIP: Fix for #1564 - using memory database in multiple threads.

blitzmann:
Technically this is a WIP while I do some more testing to make sure this works, but wanted to put it out there. This should fix the odd errors noted in #1564 where the db thread is causing errors when querying or updating the database.

See also: https://docs.sqlalchemy.org/en/13/dialects/sqlite.html#using-a-memory-database-in-multiple-threads
See also: https://stackoverflow.com/a/33057675


blitzmann:
@OzzieIsaacs This is ready for review :)

Issue #1439, opened 2020-05-27
Books synced over Kobo Sync do not store progress correctly on the device


flying-sausages:
**Describe the bug/problem**
When I read from a book synced from calibre-web, close it, and open it again, I am thrown at the beginning of the chapter. This does not happen when the same files are loaded directly.

**To Reproduce**
1. Load up [ePub](https://transfer.sh/9dUyM/The%20Teleportation%20Accident%20-%20Ned%20Beauman.epub) to calibre-web
2. Set up kobo sync
3. Download books to Kobo device
4. Flip through pages
5. Exit book reader
6. Open same book
7. See yourself at the beginning of the chapter.

**Logfile**
Kobo-side issue with the generated file, N/A

**Expected behavior**
1. Read book
2. Close book
3 Re-open book at the same place it was left on

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
![image](https://user-images.githubusercontent.com/23618693/82968546-79428e00-9fc5-11ea-9645-cced446abe1e.png)
 - Special Hardware: Kobo Forma
 - Browser: FF 68 ESR

**Additional context**
I loaded up the same file onto the reader to check if this was the case with the specific file and it works just right with the normal ePub. I'm wondering if this is something that happens to the file when it is being treated as a "Kobo ePub" as it comes from calibre-web.

flying-sausages:
Here is a [link to the file](https://transfer.sh/L0htJ/0aed9e03-f6a7-405d-bfc2-f30a54ec2c7f) that is stored in `/media/KoboReader/.kobo/kepub` which corresponds to the book linked above

LordSushiPhoenix:
I'm seeing the same behaviour on a Kobo Clara HD running the latest linuxserver/calibre-web docker on Debian Buster.

**Steps to reproduce:**

1. Add book to Calibre-web and download to device
2. Start reading and stop mid-chapter, closing the book.
3. Open same book, be at first page of chapter. 


![image](https://user-images.githubusercontent.com/19640845/83353056-d9b33180-a31d-11ea-97bf-84ad1ef3518d.png)

flying-sausages:
I've found some information on this that might help fix this.

[...]

PR #1570, opened 2020-08-17
Prefer kepub (#1439)

dmogle:
Only offer kepub download URL over Kobo sync API when available

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -344,7 +344,9 @@ def get_seriesindex(book):
 
 def get_metadata(book):
     download_urls = []
-    for book_data in book.data:
+    kepub = [data for data in book.data if data.format == 'KEPUB']
+
+    for book_data in kepub if len(kepub) > 0 else book.data:
         if book_data.format not in KOBO_FORMATS:
             continue
         for kobo_format in KOBO_FORMATS[book_data.format]:


flying-sausages:
Would it be possible to just stop offering "fake" `kepub` items altogether? This would essentially force a user to do the manual conversion, and it would reduce a lot of the confusion on why some books don't work the way an end-user would expect them to.

dmogle:
The Sync API is never offering "fake" kepubs, I think Kobo e-readers just assume that any epub offered by the API is internally formatted like a kepub. Some epubs seem to be, in my experience. We could only offer kepub formatted books but personally I would rather have a poorly formatted epub than nothing at all?

flying-sausages:
I see what you're saying. It's just that the reader will always say the format is "kobo epub" instead of just "epub". But at the end of the day each book would behave differently. On the other hand, `kepubify`ed books will all behave properly with expect-able results.

I would argue that it would be better to "push" this into a direction where it works flawlessly, and if that would involve having users convert their books (either manually or through automation) until an "on the fly" conversion is available, I would argue that would be better. Otherwise no one is pushed to improve things as they "kinda work well enough" and the users experience an "imperfect" solution.

i.e. if only `kepub`s will be allowed through on the sync, users will have a perfect experiene once they start converting books, and eventually there will be a better automation to do this if we set the experience bar this high

Just my 0.02EUR ;)

dmogle:
Hm, I get where you're coming from. On the other hand, folks who are currently using this feature might see large chunks of their library disappear from the Sync API. Maybe we should wait to see if others chime in?

flying-sausages:
Yeah this should be handled with care, I agree.

robandstuff:
I think that the risk of having users libraries suddenly be missing from the Sync API outweighs the chance to implement a "perfect" solution. Giving users a slightly better experience now and continuing to work towards an automated conversion seems like a win-win solution.

flying-sausages:
FWIW, In my own pursuit to achieve a 100% kepub library, I've gone and made this script. It's not perfect, but it's mine. https://gist.github.com/flying-sausages/2684b07e326a17f7245ca3f1812376ef

dmogle:
Thanks for posting your script, I was struggling to do something similar!

flying-sausages (approved):
tested and working

flying-sausages:
> tested and working

I might have to retract that later, I'm getting a lot of "Sync failed" on my Forma. I'm hoping it's something to do with Wi-Fi and not the simple change above. Will come back to this tomorrow.

flying-sausages:
> > tested and working
> 
> I might have to retract that later

Never mind, a factory reset fixed it all. It's working great.

flying-sausages:
@dmogle Been using this in my "prod" and so far so good. Just wanted to give you a headsup that my script had a pretty glaring flaw so all kepubs were just copies. I updated the script and added a script to find the affected books so you can find them too.

LordSushiPhoenix:
I merged this into the docker container I'm using the other day and it is working well. Books that only have epub available are downloaded as "fake kepub" and exhibit the issues that come with that, but any books that have kepub available download as proper kepubs. 

I have not tried using a script to convert all my books to kepub, just converting them as I either add new books or notice issues with previously added ones. When I notice issues, removing the download on my ereader, converting to kepub in the webui, and re-downloading is enough to get the proper kepub.

OzzieIsaacs:
Thanks for point out the problem. I found out that kobo's favorite format is epub3 format, so I just send the kepub link for epub3 format

OzzieIsaacs:
Okay, I see the problem

adamcstephens:
I bulk converted all my books with calibre, and this patch definitely fixes the problems associated with kobo syncing.

Looking forward to a new release so that the downstream docker containers get this fix.

Thanks for the fix!

Issue #1577, opened 2020-08-22
Improvements to the task queue


blitzmann:
I'm opening discussion on trying to improve the task queue. Indeed, I've actually started tinkering around with this, so I may have a PR soon, but wanted to put my thoughts out there.

While working on my own customizations for calibre-web, the lack of a plugin system has been kind of a bummer. I've resorted to planning on releasing my own fork of calibre-web with my intended feature, which should work for my purposes for now. However, just because calibre-web isn't pluggable today, doesn't mean it can't be tomorrow! One step to remedying this, I believe, would be to re-work the `Worker` thread. At the moment, all logic about tasks are integrated into the `Worker` class, and it's impossible to extend this to add new tasks if needed without modifying the literal file (something I had to do for my own customization).

So I'm planning on doing the following:
1) `Worker` class is simply a queue that runs tasks that it's given
2) Tasks are classes that are added to the worker via `worker.add()` (so, for example, a convert task would be something like `worker.add(TaskConvert(args...))`).
3) Each of these task classes extends from an Abstract Base Class that deals with most of the boiler plate stuff (each tasks should have a `run()` method, as well as started / stat / progress / etc properties)

This should open up the worker thread to be used by anything, and not just the hardcoded tasks that it comes pre-packaged with. Not to mention it keeps the code associated with the variety of tasks contained in their own class, which is good for organization and sanity 😁

Can anyone think of any issues this type of system may solve? Indeed, it's a generic job worker pattern that I've seen used in may other places, so I wouldn't think there would be any issues.

blitzmann:
This was merged in with #1580, will close :)

PR #1580, opened 2020-08-24
Refactor worker thread / tasks

blitzmann:
Issue: #1577

As mentioned in that issue, the current worker thread that handles background tasks is monolithic and non-extendable without modifying the file itself. This PR aims to address this and make the background task system more modular. 

This is a fairly big PR, but I can answer any questions and work with the maintainers to address any issues noted. Here's a quick rundown of the changes:

* The worker thread is now simply a queue that runs and manages tasks. That's its only function. 
* Each task that is added is a mostly self-contained class that extends from `CalibreTask`. This class contains the logic specific to running that task
    * `CalibreTask` is an abstract class that provides basic boilerplate logic for all tasks. This allows each task to simply focus on the stuff they need to run, while the `CalibreTask` manages things like meta data (eg: start/stop time), wrapping errors, and providing a common API that all tasks should have.
* Since each task is self contained and there isn't any reason to check the "task type" via if statements, I've removed the TASK_* constants. Each task now has their human-readable type defined in their class. This might affect localization - would be open to hearing any concerns over this and if there's an issue, how to address it
* Aside from those, the actual logic of the tasks hasn't changed much (although, I will say that I've removed the CalibreDB `add_format` task due to #1565 being merged in, which allows the database to be safely utilized within other threads)

With the changes in this PR, it should be theoretically possible for third party code to be added to calibre-web (in some sort of future plugin system) and be able to utilize the background worker thread. It simply needs to provide the queue with it's own task class extended from `CalibreTask` and it'll work like any other task

I've tested this stuff to the best of my abilities, and everything seems to be working well. I would definitely invite others to test it though.

Some things that I haven't been able to test:
* Any gdrive integration. Theoretically, it should work, but I don't have that integration set up and honestly it looked really involved that I just didn't bother. 🙄🤗
* Apparently calibre-web has tests... see `tests/`, but have no idea how to run them / if there's even tests written for these systems. So if there's any automated unit testing, they haven't been run (and indeed, they would probably have to be re-written due to the refactoring)
* I developed this on python 3.6. I think this would also work with 2.7 and the other versions that `calibre-web` supports, but I'm not positive. I don't have much experience with maintaining code that must be compatible with 2.x and 3.x, so any additional eyes on this would be beneficial


cps/db.py
```diff
@@ -338,37 +338,9 @@ def __init__(self):
...
-                    self.session.rollback()
-                    self.log.error("Database error: %s", e)
-                    # self._handleError(_(u"Database error: %(error)s.", error=e))
-                    # return
-            self.queue.task_done()
-
-
-    def stop(self):
-        self.queue.put('dummy')
-
```
blitzmann, line 371: All queue stuff was removed from CalibreDB class due to #1565 being merged in, which allows the database to be accessed safely from other threads. I've tested with adding converted formats via the Convert task, and can confirm that, at least with my many tests, adding it directly in the Worker thread works 100%

cps/helper.py
```diff
@@ -722,47 +723,30 @@ def format_runtime(runtime):
...
+                if task.stat == STAT_WAITING:
+                    ret['status'] = _(u'Waiting')
+                elif task.stat == STAT_FAIL:
+                    ret['status'] = _(u'Failed')
+                elif task.stat == STAT_STARTED:
+                    ret['status'] = _(u'Started')
+                elif task.stat == STAT_FINISH_SUCCESS:
+                    ret['status'] = _(u'Finished')
                 else:
-                    task['taskMessage'] = _(u'Unknown Task: ') + task['taskMess']
```
blitzmann, line 763: Task Type is no longer needed. Each task is it's own class and provides its own `name`. This may affect localization (should still work via `ret['taskMessage'] = "{}: {}".format(_(task.name), task.message)`) but this might be something we need to discuss

cps/services/worker.py
```diff
@@ -0,0 +1,184 @@
...
+            cls._instance = WorkerThread()
+        return cls._instance
+
+    def __init__(self):
+        threading.Thread.__init__(self)
+
+        self.dequeued = list()
+
+        self.doLock = threading.Lock()
+        self.queue = ImprovedQueue()
```
blitzmann, line 64: `queue` is a list of taskss waiting to be started. Since pythons `Queue` class automatically removes them when we call `queue.get()`, I also keep a list of tasks that were dequeued (encompassing currently running, failed, and completed tasks)

cps/services/worker.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @abc.abstractmethod
+    def run(self, worker_thread):
+        """Provides the caller some human-readable name for this class"""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def name(self):
+        """Provides the caller some human-readable name for this class"""
+        raise NotImplementedError
```
blitzmann, line 126: These abstract methods should be overridden by the concrete class. All other things here in this class simply provide default / boilerplate code for each tasks (but they can also be overridden if needed, see: `TaskEmail.progress`)

cps/services/worker.py
```diff
@@ -0,0 +1,184 @@
...
+
+#Class for all worker tasks in the background
+class WorkerThread(threading.Thread):
+    _instance = None
+
+    @classmethod
+    def getInstance(cls):
+        if cls._instance is None:
+            cls._instance = WorkerThread()
+        return cls._instance
```
blitzmann, line 56: This allows us to define `WorkerThread` as a singleton. The previous way (setting `_worker=WorkerThread()` in the file and then importing that variable) would also work, but I generally prefer this method and it makes it easier to do a few things sometimes. Let me know if there's any objections :D

OzzieIsaacs:
I didn't expect you to be so quick, there are tests in a seperate repro.
https://github.com/janeczku/calibre-web/blob/master/CONTRIBUTING.md

This tests are system tests (so an automated browser is clicking elements), so changes under the hood shouldn't influence the tests. The dark side of this is the tests are running for a long time (I think currently roughly an hour). I started testing gdrive, (I also don't use gdrive), but progress is very low.
I'll let the tests run this evening and post the outcome, and then we see.

For a big change like this it would be better to merge it to the development branch (like mentioned in the contribution rules.)

OzzieIsaacs:
grafik
Not so bad, almost half of the tests are  still passed.

[Calibre-Web TestSummary.zip](https://github.com/janeczku/calibre-web/files/5117875/Calibre-Web.TestSummary.zip)
Biggest problem seems to be:
grafik
I removed the line a restart the tests now.

blitzmann:
Ha, yeah it didn't take too much time. As I was deving out my own customization, I basically had all this in my head, just had to put it down. Went pretty smoothly, didn't really run into any troubles with it.

Thanks for the info on the tests! The calibre object not having stop - I didn't run into this during testing, probably because I simply start and stop my server via pycharm and not explicitly call that. That being said... I'm not sure CalibreDB even needs to extend from threading.Thread anymore. I will look at the tests today and see what needs to be done. 

Also, just to clarify, you want this to be merged into `development` branch? I can switch it on Github and update any merge conflicts today

OzzieIsaacs:
After rerunning the tests we are in trouble:
grafik
This looks like real problems:
[Calibre-Web TestSummary.zip](https://github.com/janeczku/calibre-web/files/5118541/Calibre-Web.TestSummary.zip)
If you replace the ..summary.html file in the test folder with this one, you should be able to see the results.
The logfiles of the tests are gone (it's on my To Do List to keep them). If one test of a class fails, it's possible that it pulls other tests to death, so some of the errors and fails are caused by the first fail of a test in the testclass.
Please check the failed tests regarding sending emails and converting ebooks, this is most probably causing most of the problem.
The testing repro is here: https://github.com/OzzieIsaacs/calibre-web-test (might be hard to get them running, I think no one tried it). But reading the tests is in my opinion pretty straigt forward, every file is deleting the app.db file and starting with a fresh instance of my testing db. The function names should explain mostly what is done, so you can reproduce the behavior.

OzzieIsaacs:
And yes, please create a pull request for merging it into development branch, and don't be afraid, should be easily mergable, as I merged it 2 day ago by myself and I didn't change anything regarding the code you created.

OzzieIsaacs:
Important for running the tests:
Deactivate attach to subprocess and deactivate gevent compatible, otherwise emailing and converting books creates trouble.
grafik

blitzmann:
Yeah, I'm trying to get the tests going in WSL, as I figured a lot of these things would be easier too set up in a linux environment. I installed calibre, java, all the requirments, etc. The instructions don't mention pointing to an existing calibre  library, so I assume the testing repo has these already set up. Then I started the tests.

image

Not sure if this is correct or not. Is it possible to check current progress? Seems like it started bu then immediately returned. 

(a Docker container with the testing environment ready to go would be amazing, btw 😋)

blitzmann:
Also, that SS notes that 11 failed, 23 errored. What are the current stats? Are the tests actually all 100% success in master / dev? What am I aiming for for break even / what tests are known failures that I wouldn't have to worry about for this PR (if any)? 

And btw, thanks for the support so far! 👍

OzzieIsaacs:
The output is not so bad, but also not good, should be something like this:
grafik
The last test run on development branch is from yesterday evening, so if you open Calibre-Web TestSummary.html in the test folder, you should get the latest stats. If you replace this file with the content of the zip file I attached in one of the last posts you should see what's going on.
This ones are definitly related to you commit:
grafik
grafik
grafik

If the rest is working, I can check for this one, as it's really ugly to debug a code which is self destroying (but I think it's also related to the change, the updater thread isn't stopped properly maybe)
grafik

OzzieIsaacs:
The test code is searching for python3.7 or python3.8 if you don't have installed one of these you won't get the pip output line and the code stops. So please add your python version or just python3, .. (very stupid of me to detect a special version) to line 43
pversion = ("python3.8", "python3.7", "python3") in main.py in the test repro, then it should work

OzzieIsaacs:
Okay I updated calibre-web-test to handle this case better and to produce an output

OzzieIsaacs:
Once the tests are running, I recommend using the pycharm debugger. Start the main.py till it has set up the virtual environment, (set a breakpoint there) and click the "stop the process" button. Then you can start every single test by hitting the green arrow beside the test function names (I guess you know this), so everything is speeding up
grafik
I don't expect you to run a complete test sequence. If you get the failing tests run provide the fixes it's fine for me.

blitzmann:
Ahh, so I'm using WSL, which is headless (the linux subsystemf or windows, basically Ubuntu in windows). Perhaps the tests require GUI. Unfortunately, I don't have a Linux desktop to test with.

I've done brief testing with Windows and the output states `Test are not guaranteed to run on windows`. After that there's an error, but I believe this might be the way I have it set up. Will keep trying to get it working :)

Thanks for the info!

OzzieIsaacs:
Sorry, now I got it, I didn‘t get WSL abbreviation. I remember running tests on windows, let me check

OzzieIsaacs:
I updated the testroutine again, it's now creating the virtual environment also on windows (I have no experience with WSL so I don't know if this better now).
The commandline for your python version should be listed in line 44 of main.py to make it run:

pversion = ["python3.7", "python3.8", "python3", "c:\\python38\\python.exe", "c:\\python37\\python.exe"]

And also the path for the python.exe in the virtual environment (I added a line which worked for my installation

blitzmann:
I think I got it to run in wsl by installing a gui and RDPing into it. I also needed to install firefox itself (not just the geckodriver) as well as change the location of ebook-convert

Might still run into issues, it literally just started and launched the browser and is currently testing against current `Develop` branch for me to get a baseline and make sure things work. Then I can look at the issues reported :D

OzzieIsaacs:
Awesome. 
Yes you need Firefox, this was clear to me, because I know what to expect. I add it up to the readme

OzzieIsaacs:
Looks my last commit to develop broke some things for the anonymous user. I pushed two commits to fix it.

blitzmann:
I've made progress on the errors, re-running the full test suite to see if that fixed a few other things as well :)

blitzmann:
@OzzieIsaacs Think I got the bugs worked out. If you want the output, I'll have to rerun the thing tomorrow - I have been selectively rerunning the tests. It should be ready for a re-review tho :)

OzzieIsaacs:
No need for running the whole suite. I‘ll check the outcome this evening, thank you

blitzmann:
Also just merged `develop` and changed the merge target 👍

blitzmann:
@OzzieIsaacs just an fyi, it looks like there may be a problem with properly deleting the tasks for the UI that have finished. At least I'm getting some funky things happening on my branch after porting over this code. I'll let you know what I find out regardless :)

blitzmann:
@OzzieIsaacs Fixed the issue with deleting old tasks. I added the logic to the underlying `CalibreWeb` tasks, so that tasks could  control when they get cleaned up (important for my customization - which is going to have tasks that dictate a minimum time to live before getting cleaned)

OzzieIsaacs:
grafik
Looks like nothing is broken.
You missed 1 or 2 strings to translate, I changed it by myself.

PR #1588, opened 2020-08-29
Improvements to task processing

blitzmann:
This is a simple and quick commit to address one specific issue with tasks processing in which once a task is removed from the queue, then processing takes place before it's added to the dequeued list. If the web interface hits between those calls, then the in-progress tasks doesn't show up. This adds the tasks to `dequeued` immediately after it's popped off the queue

It's a pretty rare edge case, but is possible, especially if tasks are finishing (or failing) quickly.


OzzieIsaacs:
There are two problems (has nothing to do with the last commit, it's general ones). 

1.) Once you started a task (like sending test e-mail) the worker thread stucks in the item = self.queue.get() line. If you now try to stop Calibre-Web, everything comes to an end, except the worker thread, which never returns. 
In my opinion the get should be implemented with timeout, so the line main_thread.is_alive is hit from time to time and allows the worker thread to stop. 
Tested on Linux Mint 20 with python3.8 and python2.7

2.) Using python2 there is a naming conflict with the "email" imports in email.py (the email... import on top of the files try to import email.py itself instead of the pyhton2 email module.) I could resolve this by renaming email.py to mail.py). If this is fine for you and you see no side effects I can do this in my repro before doing the push.

blitzmann:
Nice catch, I'll add those to this PR

blitzmann:
@OzzieIsaacs I addressed both those issues in this PR, please let me know if you see anything off. Thanks!

OzzieIsaacs:
I merged your newest version. This is working now. 

During my tests I found a ugly problem:
The test "test_convert_only" fails sometimes. Looks like some strange timing issue. 

```
        vals = self.get_convert_book(7)
        select = Select(vals['btn_from'])
        select.select_by_visible_text('MOBI')
        select = Select(vals['btn_to'])
        select.select_by_visible_text('HTMLZ')
        self.driver.find_element_by_id("btn-book-convert").click()
        self.assertTrue(self.check_element_on_page((By.ID, "flash_success")))

        self.create_user('solo', {'password': '123', 'email': 'a@b.com', 'edit_role':1})
```
The test fails after the conversion to the HTMLZ format is finished and the new user is created. Currently I don't know if this problem is caused by the code change or existing forever and due to some timing change it's now happening.
The error message is as follows:

```
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
response = self.full_dispatch_request()
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
raise value
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
rv = self.dispatch_request()
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/flask_login/utils.py", line 272, in decorated_view
return func(*args, **kwargs)
File "/home/ozzie/dev/calibre-web/cps/web.py", line 228, in inner
return f(*args, **kwargs)
File "/home/ozzie/dev/calibre-web/cps/admin.py", line 896, in new_user
languages = calibre_db.speaking_language()
File "/home/ozzie/dev/calibre-web/cps/db.py", line 650, in speaking_language
languages = self.session.query(Languages) \
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3346, in all
return list(self)
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3508, in __iter__
return self._execute_and_instances(context)
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3529, in _execute_and_instances
conn = self._get_bind_args(
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3544, in _get_bind_args
return fn(
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3523, in _connection_from_session
conn = self.session.connection(**kw)
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 1138, in connection
return self._connection_for_bind(
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 1146, in _connection_for_bind
return self.transaction._connection_for_bind(
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 409, in _connection_for_bind
self._assert_active()
File "/home/ozzie/dev/calibre-web/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 282, in _assert_active
raise sa_exc.InvalidRequestError(
sqlalchemy.exc.InvalidRequestError: This session is in 'prepared' state; no further SQL can be emitted within this transaction.
```

I tried to run it with the old version and it always works there.

blitzmann:
Thanks for letting me know.

Can I got some more information on the `test_convert_only` test? What is it supposed to do? I don't have the project in front of me at the moment (at work). Is it converting a book and then emailing it to a user? Why is it creating a new user for a convert test? Just want to make sure I have a better understanding of the purpose of the test before jumping in tonight :)

OzzieIsaacs:
Yeah, you got it

OzzieIsaacs:
If you try to run the tests from the repro, try the development branch of the calibre-web-tests. I fixed several testing problems

OzzieIsaacs:
I read the old code and now I remember: This spot in the code was handled by giving the parameters to write to the database thread, because something similar to now failed before and I could not find the problem.

blitzmann:
@OzzieIsaacs Got some time today to look into this, I cannot for the life of me get this test to fail. I've run it 10 times now, all passing. How often was it failing for you?

Just to clarify, I'm using the development branch of calibre-web-test (df3e2e) and my tasks branch (ef49e2).

Was it failing for you in the same spot or did it seem to jump around a bit? According to the traceback it's failing during the creation of the new user when trying to get languages from the database. Are you able to reproduce the error if you disable most of the other converts (to maybe help speed up testing of it?)? 

I haven't looked into the exception itself or tried to trace it back into trying to logically deduce the issue, just been trying to get the test to fail at least once. :(

OzzieIsaacs:
With my "Linux testing computer" an older one, the test fails every 2nd run. Currently I'm running the test on my laptop (also Linux) and it looks indeed it's not failing there (4 in a row without problems). 
Just to clarify, I'm using the development branch of calibre-web-test (df3e2e) and my tasks branch (ef49e2). -> Yes
Was it failing for you in the same spot or did it seem to jump around a bit? -> Was always the same line, the "speaking_language" function. From looking into the code it looks to me like this call is the first call to the session in the main thread after adding the book in the worker thread. 
Are you able to reproduce the error if you disable most of the other converts (to maybe help speed up testing of it?)? Havend tried this, but I guess yes. I googled a bit for the error message and the usual recommendation is to have session objects per thread (https://stackoverflow.com/questions/39126810/this-session-is-in-prepared-state-no-further-error-with-sqlalchemy-using-sc).
(alsso, I'm using python 3.6 for the tests, whereas your traceback shows 3.8. I may have to try with that version to see if I can reproduce) -> I don't think it makes a difference.  As mentioned above I had problems at this spot (due to the only spot in the code where the calibre database is write accessed not from the main thread) and therefore generated this queued thread communication code from before.
I tried some small changes like activating autocommit, issuing a session flush before doing the commit in the worker thread, but didn't change anything. 
I wanted to chnage to code so the worker thread has it's own session but was more than a one line, so I didn't do it yet.

OzzieIsaacs:
On my 2nd computer it took 6 runs to get the error. I try if I can remove all the other conversion steps to speed up the tests

OzzieIsaacs:
I found something working for me (even if not understood):
First I made the Session variable in db.py global and created a session() in the main thread and one in the  convert thread. Before Both sessions were point to the same session. After changing everything in convert to the convert thread session I had some data consistency problems (the new added format wasn't visible in the books details setting but the book was listed in the formats nav-section. So I tried to remove the Session after the commit in the convert thread and got an error message (`sqlalchemy.orm.exc.DetachedInstanceError: Instance <Books at 0x7fbe84e305e0> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/13/bhk3)`). This led me not attach the new_format to current_book.data but only adding new_format to the session and activating autoflush. This made this error also disappear. I ran this patch over 40times on my slow computer which was showing the problem more often. The patch is now merged to a new branch database_fix. 
My final idea is to revert the new session in the thread stuff and only change the "add to book routine" to the procedure described above and check if nothing else is broken.

OzzieIsaacs:
I reverted the "thread local session", the error came back after 3 runs, so my conclusion is: we need this local session in the thread. Everything else seems to work.
If it works for you I would merge it back to the development branch

blitzmann:
@OzzieIsaacs sorry for my delayed response. I got the testing environment up and running on an oldish laptop, but after 20+ runs of that test I still couldn't get it to fail.

I believe the change you're talking about is feacbe8ebd35803ebd9ed88cc931564cfc9bcb86. It makes sense to me.  After looking at the SQLAlchemy docs, I do believe this is actually the proper use of `scoped_session`. Instead of making a scoped session, but then running everything through the same session object that ti produces, we need to make available the factory that `scoped_session` produces so that other threads can access it and make their own session.

So that change does make sense 👍 I'm going to defer to your judgement on final tweaks that need to be made since I can't seem to reproduce this issue at all, but if that fixes the issue and allows the tests to pass consistently, I say go for it 😄

OzzieIsaacs:
At least it's not breaking things for you. I merged it to the development branch. 
It will take some time for merging it finally into the master branch

Issue #1484, opened 2020-06-11
It is not possible to save notes and highlights in books downloaded using kobo sync


AntonioDeTeo:
**Describe the bug/problem**
Hello!
I've been trying kobo sync for a few days and it's great, thanks for your work.
I noticed that epub books downloaded using kobo sync do not allow to save highlights and notes. While testing, I have seen that if a book is only available in kepub format, it downloads that format and those features do work perfectly. If the book is in kepub format but there is also an epub format, it seems that it is always the epub version that is downloaded, so the problem persists.
To solve this problem, I would propose that if there is a kepub version of the book, kobo sync downloads that one instead of the epub version.

1. Download a book stored in epub format using kobo sync
2. Open the book and try to save a note
3. Go to the book notes and check that it hasn't been saved 

**Expected behavior**
If there is a kepub version of the book, it should be downloaded instead of epub to solve these problems.
If there's a way to make these features work directly from an epub it would be great, because it would save me from having to convert to kepub all the books in my library. But I understand that it would be easier to give preference to kepubs for downloads.

My Environment:
 - OS: FreeNAS-11.3-U3.1
 - Python version: python3.7
 - Calibre-Web version: 8/6/20 21:56
 - Docker container [ None/Technosoft2000/Linuxuser]:
 - Tested in a Kobo libra and a Kobo Forma

OzzieIsaacs:
I fixed a bug in storing the progress, so if you update to the newest commit version the bug might be fixed

AntonioDeTeo:
Thanks for the fix, but the problem persists.
The bug I had detected was not related to the synchronization in the cloud. The problem occurs with any book in epub format that I download using kobosync: once downloaded, I go to any page and exit. If I enter again, the lecture resumes at the beginning of the chapter and not on the page where I had left it. As I mentioned in the previous comment, saving notes or highlights doesn't work either.
This only happens with epubs downloaded via kobosync, if I import the same epub to my device using an usb cable, everything seems to work fine. As I also commented before, if I force the book to be downloaded via kobosync in kepub format (leaving only the kepub version in calibre-web) everything works fine too.
It seems that the kobo device expects everything downloaded from its online store (or from kobosync in this case) to be in a real kepub format and uses some different system to save the reading progress, highlights and notes, but the ebook is really just an epub with the extension changed (or something like that) and something is wrong for that reason.
[...]

Issue #1488, opened 2020-06-11
On-the-fly conversion for Send to Kindle and Kobo sync


flying-sausages:
**Is your feature request related to a problem? Please describe.**
Currently, there are issues with Kobo Sync not necessarily serving up the kepub when an epub is present (#1439), and from what I gather, books need to be in either azw3 or mobi format in order to be sent to a Kindle. This means that a bunch of manual steps are needed before books can arrive on a reader.

**Describe the solution you'd like**
It would be nice to have the necessary conversions done for the platform at hand, when the sync/sending is being performed. This would alleviate users from having to resort to manually creating the right versions of their books en-masse which is currently only really possible through Calibre, and a lot of confusion about why books behave in odd ways even though the format is "correct" on the surface.

**Describe alternatives you've considered**
- En-masse batch editing through Calibre GUI
- Making a wrapper script to create mobi and kepub versions when they are being added to calibredb

**Additional context**
Converting large books might be an issue with the Kobo Sync as it might time out, however I have not tested this in any way.

LordSushiPhoenix:
If converting large books ends up making the sync time out, maybe calibre-web could automatically run kepubify when new books are added if kobo sync is enabled? Then when syncing it would need to prefer the kepub over the original epub. 

Even just syncing kepub instead of original epub if kepub is available seems acceptable to me.

flying-sausages:
> If converting large books ends up making the sync time out, maybe calibre-web could automatically run kepubify when new books are added if kobo sync is enabled? Then when syncing it would need to prefer the kepub over the original epub.
> 
> Even just syncing kepub instead of original epub if kepub is available seems acceptable to me.

That might be undesirable with large libraries. Also, that would be a bit annoying to handle for books added through `calibredb add` as the library will then have to be monitored for that. I would accept that as a solution too I suppose.

flying-sausages:
I'm actually not sure if this should be a bug or not as the README.md mentions this should be functional, however in the case of kobo it is not.

>Optionally, to enable on-the-fly conversion from one ebook format to another when using the send-to-kindle feature, or during editing of ebooks metadata:
>
>Download and install the Calibre desktop program for your platform and enter the folder including program name (normally /opt/calibre/ebook-convert, or C:\Program Files\calibre\ebook-convert.exe) in the field "calibre's converter tool" on the setup page.
>
[...]

PR #1596, opened 2020-09-01
Automatic conversion in kepub during Kobo sync

norangebit:
The Kobo e-readers assume that ebook downloaded during sync are in `KEPUB` format, so they interpret as `KEPUB` even if they are `EPUB`.
This misunderstanding generates strange malfunctions (#1439 & #1484) that affect the reading experience.

The PR #1570 prefers the `KEPUB` format over the `EPUB`, but only if this is available, so the user must perform a manual conversion before starting the synchronization in order to have a correct behavior.
You would have a better user experience, and less error prone, with automatic conversion.

In my opinion the best time to perform this conversion is during synchronization.
In fact, in this way, ebooks are converted only when conversion is strictly necessary and it is ensured that all ebooks synchronized with the Kobo are actually in `KEPUB`.

This PR converts all new ebooks, only if they are not available in `KEPUB` and the kepubify path has been set.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -178,6 +178,10 @@ def HandleSyncRequest():
 
     reading_states_in_new_entitlements = []
     for book in changed_entries:
+        formats = [data.format for data in book.data]
+        if not 'KEPUB' in formats and config.config_kepubifypath and 'EPUB' in formats:
+            helper.convert_book_format(book.id, config.config_calibre_dir, 'EPUB', 'KEPUB', current_user.nickname)
+
         kobo_reading_state = get_or_create_reading_state(book.id)
         entitlement = {
             "BookEntitlement": create_book_entitlement(book, archived=(book.id in archived_book_ids)),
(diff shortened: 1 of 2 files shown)


norangebit:
In #1488 there's mention of possible time out problems during synchronization in case of very large libraries.
A possible solution could be to convert the entire library to `KEPUB` when creating the token for synchronization and use the conversion during synchronization only for new ebooks.

This solution ensures that all books downloaded to the Kobo are in `KEPUB` so to avoid the above mentioned problems and should also avoid time out problems.
On the other hand it requires the conversion of the entire library, but if you want to read on the Kobo this is a necessary evil.

flying-sausages:
@norangebit FYI there's some merge conflicts that popped up with the merge of #1570

OzzieIsaacs:
The conversion itself isn't done "inline", it's done in background in another thread so there might be a problem if the sync is ongoing and in parallel the database is changing as kepub's are "entering" the database

Issue #1466, opened 2020-06-04
Logout after search in Browser on Tolino E-Book reader


Gagalari:
I login into calibre-web on my reader. this works fine. but when I search a book or click any other button, i get logout. This issue I only have on reader. On any other device like phone or laptops I don't have this problem.
Any suggestion to fix this problem or is this may be a bug?

**Logfile**
In the log file there is nothing, only calibre-web was starting in the morning. No crash nothing.

**Expected behavior**
Like this, the calibre-web on my readers (all are tolinos) are useless, because after every search I must login and my search is away. 

**Environment (please complete the following information):**
 - Client: Tolino - E-Reader
 - Python version 3.7.3 (default, Dec 20 2019, 18:57:59) [GCC 8.3.0]
 - Calibre-Web version 0.6.9 Beta - b83d56eff2c820e55222c9856000f0422f721f3d - 2020-06-03T19:10:36+02:00
 - Docker container None
 - Special Hardware Raspberry Pi 3
 - Browser browser from tolino E-Reader

Gagalari:
> I don't have access to a tolino. I checked it with my kobo, it's working there.
> Did you use Calibre-web before? (with older versions)

No it's first time I use calibre web. it's so nice on every device it works.
I don't have another type of reader here, only tolinos, and I think all with same software-version on it.

it's really special always when I press a button in menu or search I will forward to a url like this:
http://ip:8083/login?next=%2Fauthor%3data%3Dauthor%26sort%3Dnew (example when I press on author).

OzzieIsaacs:
You are right this is the link to the login page and the "next=" part shows that you are pushed on authors before.
I remember somebody saying tolino (or kobo?) behalf sometimes very strange if not used on standard ports. So from my point of view it might be worth trying to use a reverse proxy, or changing the port for testing to port 80

Coppertick:
Hi both. Having the same issue with A Tolino Vision 5 (Kobo Libra H2o). And I can confirm that it did work with an older version I had installed as "stand alone" on another server some months ago. And it also did work then with a port != 80.

OzzieIsaacs:
We did some security hardening, maybe this is causing it:
Could you try to delete in cps/web.py line 134ff:
```
@app.after_request
def add_security_headers(resp):
    # resp.headers['Content-Security-Policy']= "script-src 'self' https://www.googleapis.com https://api.douban.com https://comicvine.gamespot.com;"
[...]

PR #1608, opened 2020-09-08
Implemented HSTS config via web frontend (fixes #1607)

### LINKED ISSUE #1607: Enabling/Disabling HTTP Strict Transport Security via web frontend
Enabling HSTS requires changing the source code;  un-commenting line https://github.com/janeczku/calibre-web/blob/1450a21d0002faa39f3c07228139113c9a7b36ed/cps/web.py#L135

Enabling this security feature is a matter of configuration, so a checkbox in `admin/config` ("Basic Configuration -> Server Configuration") would be the correct way.

Alternatively one could edit the source code as described above. But depending on the method of updating this change can be overwritten by upstream changes affecting `web.py`.

**Additional context**

timholl:
This PR implements the feature to enable/disable HSTS without changing the source code. See #1607

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -648,6 +648,9 @@ def _configuration_update_helper():
         _config_string(to_save, "config_converterpath")
         _config_string(to_save, "config_kepubifypath")
 
+        _config_checkbox_int(to_save, "config_enable_hsts")
+        reboot_required |= _config_checkbox_int(to_save, "config_enable_hsts")
+
         reboot_required |= _config_int(to_save, "config_login_type")
 
         #LDAP configurator,
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -128,6 +128,8 @@ class _Settings(_Base):
     config_reverse_proxy_login_header_name = Column(String)
     config_allow_reverse_proxy_header_login = Column(Boolean, default=False)
 
+    config_enable_hsts = Column(Boolean, default=False)
+
     def __repr__(self):
         return self.__class__.__name__
 
--- a/cps/templates/config_edit.html
+++ b/cps/templates/config_edit.html
@@ -105,6 +105,10 @@
           </span>
         </div>
         <div class="form-group">
+          <input type="checkbox" id="config_enable_hsts" name="config_enable_hsts" {% if config.config_enable_hsts %}checked{% endif %}>
+          <label for="config_enable_hsts">{{_('Enable HSTS (for SSL Servers only)')}}</label>
+        </div>
+        <div class="form-group">
           <label for="config_updatechannel">{{_('Update Channel')}}</label>
             <select name="config_updatechannel" id="config_updatechannel" class="form-control">
               <option value="0" {% if config.config_updatechannel == 0 %}selected{% endif %}>{{_('Stable')}}</option>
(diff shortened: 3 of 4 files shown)


OzzieIsaacs:
Thanks for the fix. My idea would be more like detect the Tolino reader and deactivate the function transparent for the user, as any other browser would benefit from this feature. If you have a setting, like you implemented it, this would  lead to have it deactivated for anybody.

timholl:
The point of HSTS is to prevent MITM attacks by enforcing encrypted connections to all clients; why should there be exceptions for certain clients? Is the Tolino reader unable to use HTTPS in general or does it just not support the HSTS header?
I would say that if someone wants to use a reader that is not capable of a certain security feature, it should simply not be enabled in the settings in the first place.

OzzieIsaacs:
> Is the Tolino reader unable to use HTTPS in general or does it just not support the HSTS header?

Yes, it looks like Tolino reader isn't able to handle this header. This was the reason why the code was commented out

OzzieIsaacs:
Issue for your reference: #1466

timholl:
There are two options:

 1. Keep the global option to enable HSTS and add a label to it like „Devices which do not support this security feature (like Tolino readers) will not be able to connect.
 2. Make exceptions for those readers. This requires that those readers can be detected, like by their `User-Agent` header. Since I would not like to hardcode and maintain a list of those reader's UA headers, this should be configurable by the user. My suggestion in that case would be a user-defined regular expression.

timholl:
Seems like this PR is now obsolet by https://github.com/janeczku/calibre-web/commit/f06cc25a990207431250197abadda101b38b8198#diff-be8c7650699ca7209b1492ab8a139f6eR135 .

> Since I would not like to hardcode and maintain a list of those reader's UA headers, this should be configurable by the user.

The now implemented solution hardcodes an exception for readers with `tolino` in their UA header.
There is no way to prevent this exception or extend it for other outdated readers (without editing the source code).

Also, HSTS headers will now be sent ever, even on non-HTTPS connections and for people without properly configured TLS certificates and local domain names, not sure which impacts this might have.

OzzieIsaacs:
The setting was activated in an earlier version and some users of Tolino readers complained about it, nobody else ever reported an issue. 
I see this currently as a test if it causes a problem for somebody else. If somebody else has problems with this setting I consider using your suggestion. Currently I try to keep the UI as simple as possible with as less as possible settings regarding server configuration (which is hard enough)

OzzieIsaacs:
The tolino browser seemsto be very bad implemented. I'll doublecheck the setting with my kobo reader

OzzieIsaacs:
Sorry I got confused, the tolino problem was something else (the session "strong" stuff). I can't remember why I commented this lines out. I have to search for it, there are 2 posibility:
1) I thought it was to risky right activating it right before a release and deactivated it
2) there was a problem with something

OzzieIsaacs:
> Also, HSTS headers will now be sent ever, even on non-HTTPS connections and for people without properly configured TLS certificates and local domain names, not sure which impacts this might have.
Mozilla pages states the following:

grafik


So it looks to me like doing it always is not so bad.

xiazeyu:
Just some issues on my use case here.
HSTS is a good feature, while I use a mix of HTTP and HTTPS solutions on my single domain, seperated ports.
For those HTTPS sites, it works fine, but for thoese HTTP sites shared a same domain but different port, the browser will always be redirected to a non-existed HTTPS protocal. So I have to go to `chrome://net-internals/#hsts` and mannually deleted the HSTS policy everytime.

PR #1613, opened 2020-09-11
Parse PDF XMP metadata on upload

rscmbbng:
This PR fills metadata fields based on available XMP metadata present in a given PDF. 

It checks whether there is XMP metadata and otherwise it falls back to the existing  `getDocumentInfo` method. ( l:120-l:143)

Additionally, language tags embedded in the metadata are now checked against `isoLanguages` so they can also be automatically displayed. (l:144-153)

Finally, the first image of the PDF is extracted as a cover even when there is no PDF metadata. (l:109)

XMP metadata is added to PDFs when using Calibre's various 'Embed Metadata' functions. This is handy because it makes the metadata of the files portable and doesn't require the work of double entry (first in Calibre, then in Calibre-Web).


OzzieIsaacs:
Sorry for the late reply. 
Meanwhile I had some time to do some tests and I found some problems:
[invalid.zip](https://github.com/janeczku/calibre-web/files/6136438/invalid.zip)
If you are uploading the (invalid) _book1.pdf the code aborts due to a attribute error
with the (valid) book1.pdf I get a "German" is an invalid language warning upon uploading it

rscmbbng:
Hi @OzzieIsaacs, thanks for checking. I addressed the issues you've found and added exception handling as well. I've split off the XMP parsing logic into a separate function. As it is now it will try XMP, then look for DocInfo metadata and otherwise fallback to the defaults.

As it stands now only the first language is recognized in documents with multiple languages assigned. I believe that is because of `edit_book_languages` is called upon uploading. That is a separate issue though.

PR #1620, opened 2020-09-19
Remove global Session object, replace with CalibreDB class

blitzmann:
Recently a global session object was introduced (feacbe8ebd35803ebd9ed88cc931564cfc9bcb86) that allowed another thread to get a local session object for it's own use.

I had a need to not just get a new session, but also utilize functions that come with the CalibreDB class. I couldn't just call the global `calibre_db` variable tho because it contained a session that was incompatible with other threads, and I didn't want to rewrite all the methods to have a session object injected as an argument

This PR aims to improve upon that. the database setup is now done on the class itself. Local sessions can be created using `local_ssession = CalibreDB()`. The `__init__()` function will use the previously set up session maker and set `self.session`. Additionally, it will register itself to the class's ` instances` list - this is required so that we can iterate over all active instances and dispose of them correctly (that being said, there will usually only ever be 1 instance unless a thread / task has created another).

I should note that I ran this through the testing framework and everything seems like it's working :)

### DIFF
--- a/cps/__init__.py
+++ b/cps/__init__.py
@@ -83,6 +83,8 @@ log = logger.create()
 
 from . import services
 
+db.CalibreDB.setup_db(config, cli.settingspath)
+
 calibre_db = db.CalibreDB()
 
 def create_app():
@@ -91,7 +93,7 @@ def create_app():
     if sys.version_info < (3, 0):
         app.static_folder = app.static_folder.decode('utf-8')
         app.root_path = app.root_path.decode('utf-8')
-        app.instance_path = app.instance_path .decode('utf-8')
+        app.instance_path = app.instance_path.decode('utf-8')
 
     cache_buster.init_cache_busting(app)
 
@@ -101,7 +103,6 @@ def create_app():
     app.secret_key = os.getenv('SECRET_KEY', config_sql.get_flask_session_key(ub.session))
 
     web_server.init_app(app, config)
-    calibre_db.setup_db(config, cli.settingspath)
 
     babel.init_app(app)
     _BABEL_TRANSLATIONS.update(str(item) for item in babel.list_translations())
--- a/cps/db.py
+++ b/cps/db.py
@@ -42,8 +42,11 @@ from flask_babel import gettext as _
 from . import logger, ub, isoLanguages
 from .pagination import Pagination
 
+from weakref import WeakSet
+
 try:
     import unidecode
+
     use_unidecode = True
 except ImportError:
     use_unidecode = False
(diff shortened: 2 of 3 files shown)


blitzmann:
@OzzieIsaacs any feedback on this? I've got another PR ready that is kind of based on this  but want to make sure this change looks okay before finalizing it :D

blitzmann:
Okay cool! Let me know if any of the tests fail and I can take a look

The next PR (and probably last for a while) that I have coming up might be a bit controversial. It's splitting out the logic of adding a book from the Flask route to it's own function that takes in book meta object as well as a session object to use. My customizations needed it separate as I add a book from a Task thread and right now it's integrated into the route directly which isn't available in other threads. I saw it might be controversial simply because there's multiple ways of tackling the problem - I did it in a simple way which should reduce any side effects, but there's "more proper" ways of doing it. I can detail my thoughts more clearly when I open the PR once this PR is tested / accepted. Only reason I bring it up is because, if at all possible, I'd like to get the next PR into Develop before a merge into master / release, because that will allow me to finalize my own customization :D

OzzieIsaacs:
Hmm:
```
[2020-09-29 19:20:11,315] ERROR {cps.services.worker:164} 'CalibreDB' object has no attribute 'merge'
  File "/home/ozzie/Entwicklung/calibre-web/cps/services/worker.py", line 161, in start
    self.run(*args)
  File "/home/ozzie/Entwicklung/calibre-web/cps/tasks/convert.py", line 34, in run
    filename = self._convert_ebook_format()
  File "/home/ozzie/Entwicklung/calibre-web/cps/tasks/convert.py", line 97, in _convert_ebook_format
    local_session.merge(new_format)
AttributeError: 'CalibreDB' object has no attribute 'merge'
```

To reproduce: Start Calibre-Web (e.g. with test-db) login as admin, try to convert book from epub to lrf (I tried it with book7), the task fails (visible in the task list) the logfile shows the above traceback.
Checked with python2.7 and python3.8 on linux Mint

OzzieIsaacs:
Testcase `test_convert_only` fails, all conversion fail, this is detected by the test.

OzzieIsaacs:
Beside this problem all tests seem to run fine:
grafik
(The red line marks the tests which are failed because of this problem, it looks like it's always the same one)

OzzieIsaacs:
I did some more tests and found a problem from the past. The old behaviour was creating 2 tasks in the list for converting and mailing a converted book:
grafik
Now it creates one task in the tasklist:
grafik

This is because the email task is created by bypassing the "add" Task mechanism and just starting the task. So you can generate an overall failed/finished status for both tasks. But in my opinion it would be better to have 2 separate entries, because if the convert succeeds the book is in the database, also if the email task fails. 
What is your opinion (did you have some special usecase in mind by changing it) ?

blitzmann:
I'll double check the failing test this weekend :D

As for the convert / email tasks, just clarifying that that isn't part of this PR, but rather is due to some of the recent changes I've made with tasks in general, correct? Or do you think this is the reason for the test failure?

I don't have any specific reason to not have both tasks show up as explicit tasks. honestly, this is probably more of an oversight than anything. I can take a look at tweaking this as well :)

OzzieIsaacs:
The failed tests is because of the stacktrace you see in this post: https://github.com/janeczku/calibre-web/pull/1620#issuecomment-700864470 (so if one of the failed tests works, everything works, you could just start a conversion of a book with calibre and you see the problem).
Regarding separate tasks for email and convert, I created a fix and published it in "worker_task" branch, please have a look if you have a better idea to fix it, of if it's fine for you. This is nothing new, it's from the original code change you made (I just didn't realize it, and the tests weren't checking the content of the tasks results detailed enough). It has nothing to do with the failed tests.

blitzmann:
@OzzieIsaacs made a commit that should fix the convert issue. I had `local_session` pointing to the incorrect thing

As for 23d66a0d68ece521b61e4633a3694dfcef765225, looks totally fine and seems to work well. 👍

OzzieIsaacs:
Thanks, that did the trick

Issue #981, opened 2019-07-27
Add support for infinite scroll with Caliblur theme


gpapin:
Everything is in the title pretty much...

If you have the Caliblur dark theme activated and click the "recently added" link then the only way to see more is to click the arrow at the top right corner.

OzzieIsaacs:
I think this is intentional.
The Caliblur! Theme was developed by @leram84, he did all the design decision

leram84:
So, yes it was a design choice, but kind of one of necessity more than preference. Im actually open to the idea of infinite scroll if people would prefer it, but this is something i couldn't figure out. There is a js function that calls the page to load more content, but i can't figure out what wrapper its attached to. Something about the way i reordered the page is causing that wrapper to not get triggered and i don't know the js to add it to the caliblur page wrapper. If anyone can figure it out, and if ozzie approves, ill change it, but I honestly have no idea how to make it happen.

gpapin:
Ok good to know you are both open to it, I may be able to help if I find some time.

leram84:
@gpapin were you able to look at this? I'm working on integrating the newer features, and if you end up adding this, i can redesign the sorting options to take the place of the page arrows.

gpapin:
No not yet, I have to get it merge though. It's almost there I think. If you look at https://github.com/janeczku/calibre-web/pull/986 you'll see that @OzzieIsaacs has presumably found the source of the issue so I gotta find some time to try it out. If you refer to infinite scrolling though I didn't get any change to look at it at all.

Syllius24:
BUMP! Please keep this thread going, I'm incredibly interested in infinite scroll on the content server.

Wdavery:
Another vote here for this, would love to see scrolling implemented!

ghighi3f:
Is someone working on this one ?
I think I found the issue, and also one possible fix.
From the [Infinite Scroll Docs](https://infinite-scroll.com/options.html#scrollthreshold)
_Sets the distance between the viewport to scroll area for scrollThreshold event to be triggered._
This threshold was never met with CaliBlur Theme on, because the scroll is done inside a div and that doesn't trigger the Viewport scroll event. 

Also I've commented a set of rules for calibreBlur display of each book.
In my tests, after loading new pages, they conflict with isotope and the results was buggy.
Anyway, they look almost the same for me without those rules, and the InfiniteScroll works as expected.

PR #1655, opened 2020-10-10
Issue/caliblur 981 - Infinite Scroll on caliBlur Theme

ghighi3f:
Work on #981

### DIFF
--- a/cps/static/css/caliBlur.css
+++ b/cps/static/css/caliBlur.css
@@ -637,18 +637,18 @@ div[aria-label="Edit/Delete book"] > .btn > span:hover {
     color: var(--color-primary)
 }
 
-.book {
+/* .book {
     width: 225px;
     max-width: 225px;
     position: relative !important;
-    left: auto !important;
+    left: auto !important; 
     top: auto !important;
     -webkit-transform: none !important;
     -ms-transform: none !important;
     transform: none !important;
     min-width: 225px;
     display: block
-}
+} */
 
 #infscr-loading img, body > div.container-fluid > div > div.col-sm-10 > div.discover > div.isotope:after, body > div.container-fluid > div > div.col-sm-10 > div.discover > div.isotope:before {
     display: none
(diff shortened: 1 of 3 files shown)


OzzieIsaacs:
The Infinite Scroll now works, but I think the behavior is a bit strange. 
In Default Theme it's starting to scroll infinite or it's displaying the paging buttons
In the Caliblurtheme, it displays the paging buttons in the upper right corner and then you can get it to scroll infinite. If you now press on the '->' Button you get the content of the 2nd page (no matter how many books you've been loaded before)
I this will maybe time consuming to change, what do you think. 
(I'm not using this theme, so it's fine for me, but I know my users, the next issue regarding this is almost filed..)

ghighi3f:
Ok, I get where you're coming from. If user starts at the page 1 and scrolls down all the pages, when he scrolls back to the top, he could press the next button expecting a new page, not the second one.
One difference between themes is that on Default the pagination is placed at the bottom of the page, and it hides when new books are added to view.
Until the last page, the user cannot click the pagination, if I'm correct.
Also, the pagination has page numbers on Default, so user can choose a specific page.
On caliBlur, the pagination is placed on top, so user can click it without scrolling, or after he scrolls back to top, and there are no page number links, only Next '->' or Prev '<-' if available.

I'm thinking at 2 variants:

> now that the Infinite Scroll is working on CaliBlur Theme, the next and previous arrows could be hidden, like on the Default Theme

> or we should update the Infinite Scroll script so it changes the href of those arrow anchors accordingly to the last page's url that user have been scrolling to (because the url is already changed dynamic when the user get a new page by scrolling, both on caliBlur and Default)

I'm not fully sure about the second solution tho, because to press on those arrows the user have to scroll back to the top, so I don't know if the page that he was scrolling to until he stopped is still relevant ❓ 

I was focusing on fixing the Infinite Scroll, I agree with you that l the new behaviour can be a little confusing.

Let me know what you think, I'm open to implement any of the solutions user of the theme would agree on.

ghighi3f:
Halo.
I was overthinking the problem and couldn't see one obvious solution.
So now when scrolling to bottom this should happen:

- books from the next page are added to view

- window.history state is replaced

- the .pagination div is replaced also with the next page's one, so it stays in sync

Lemme know
cheers

PR #1659, opened 2020-10-11
fix - restore read search in advanced tab

celogeek:
The read search wasn't active. I fix the call to get_cc_columns.

The only place when the Read CC show be hidden is for Book Preview.

PR #1662, opened 2020-10-12
Bubble read status

celogeek:
This is the render.
I need to handle the read status for books.

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -819,3 +819,18 @@ def get_download_link(book_id, book_format, client):
         return do_download_file(book, book_format, client, data1, headers)
     else:
         abort(404)
+
+
+def get_readbooks_ids():
+    if not config.config_read_column:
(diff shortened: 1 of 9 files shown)


celogeek:
This PR includes https://github.com/janeczku/calibre-web/pull/1659

celogeek:
Can you tell me what you think about the style?

Capture d’écran 2020-10-12 à 09 35 49

celogeek:
I try to figure out an easy and global way to add to the Book the read status. All of the views depend on it.
If you got an idea, I'm wide open ;)

celogeek:
I update to match the bubble style:

Capture d’écran 2020-10-17 à 13 16 10

Pending: apply badge only when read

celogeek:
Bubble read is ready. Can you check it please and let me known if the patch is ok?

Issue #1428, opened 2020-05-24
CBZ files do not have a convert option available


flying-sausages:
**Describe the bug/problem**
When editing metadata for CBZ files, the converting function is not available

**To Reproduce**
1. Add CBZ file to library
2. Go to file's metadata field
3. Find a lack of converting options

**Logfile**
N/A

**Expected behavior**
Being able to use the Calibre converter to create ePub/PDF etc. versions of the file

**Screenshots**
![image](https://user-images.githubusercontent.com/23618693/82755778-c9570000-9dcd-11ea-8a31-75571345b621.png)

![image](https://user-images.githubusercontent.com/23618693/82755794-df64c080-9dcd-11ea-92f6-3304ceb70184.png)

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Python version: Python 3.8.2
 - Calibre-Web version 0.6.8:
 - Docker container [ None/Technosoft2000/Linuxuser]: None
 - Browser [e.g. chrome, safari] : Firefox ESR 68

**Additional context**
While the "ebook converter" is shown as "not installed", the option is available for epub and mobi and probably most other formats
![image](https://user-images.githubusercontent.com/23618693/82755833-2357c580-9dce-11ea-9865-570ede057bb5.png)

flying-sausages:
Fixed the path of "ebook converter" to the calibre one and tested succesful conversion, but the issue still persists

OzzieIsaacs:
From my knowledge the calibre-converter only supports a limited subset of converting files to another format. 
I think all the comic file formats (like .cbz, .cbr, .cbt) are not supported formats as source or destination format

flying-sausages:
I got succesful, nice and quite optimal converts with `ebook-convert` just now. Just need to pass a couple flags to the converter

flying-sausages:
i.e. https://gist.github.com/flying-sausages/867021e2ee1ee5aa31d8356a3dbc90f9

OzzieIsaacs:
You are right, the algorithm for showing the conversion options has to be more sophisticated. Currently it is "one" list for all source and destination format. But Calibre can convert ''from'' more formats than **to** formats.

flying-sausages:
I would think the easiest way would be to have the database store settings like these
```
Rule NR | FROM format   | TO format     | Flags to add               |            
------------------------------------------------------------------
[...]

PR #1667, opened 2020-10-16
Fixed a few minor issues with display and added the ability to convert cbr/cbz (so that you can send comics to kindle)

cbartondock:
- Fixed issue caused by new "series" metadata where not all grid entries would be filled. Fixed by using flex display, which is just better anyways.
Capture3

- Fixed cover images having incorrect widths in the Discover page. 
Capture2

- Added split "EXTENSIONS_CONVERT" constant into "EXTENSIONS_CONVERT_FROM" and "EXTENSIONS_CONVERT_TO" to allow converting from CBZ and CBR to mobi (for example)
Capturea


OzzieIsaacs:
Addresses #1428

cbartondock:
Yep. Let me know if you need anything changed prior to merge.

cbartondock:
Any reason not to merge?

PR #1684, opened 2020-10-27
Increase the size of the generated auth token

jvoisin:


OzzieIsaacs:
You don't want to enter manually this code, don't you:
new_auth
Instead of this one:
link
What we would need to make this work would be having a page which presents you this link and you can just follow the link to authenticate your other device. I will not merge this one

Issue #1693, opened 2020-10-31
KoboSync Download Links


naydichev:
**Describe the bug/problem**

- first of all, my Kobo sync isn't working at all, but there is an existing issue open about this (#1680).
- I tested the call locally (using `curl`) and I noticed that there are download links that do not go through the kobo routes. I have explicitly opened up `<calibre-web>/kobo` to be accessible by the outside world, all other paths require SSL MA. I'm not aware of a mechanism to setup SSL MA with the Kobo reader, but I do believe that all traffic from Kobo should be routed through a `/kobo/<auth_token>/` URL. It doesn't seem like it would be too difficult to make this change either.

**To Reproduce**
1. `curl $CALIBRE_URL/kobo/$KOBO_AUTH_TOKEN/v1/library/sync`
2. Notice the `URL`: `"Url":"$CALIBRE_URL/download/84/epub"`
3. It should be `"Url":"$CALIBRE_URL/kobo/$KOBO_AUTH_TOKEN/download/84/epub"`

**Expected behavior**
See `#3` from the list above.

**Screenshots**
N/A

**Environment (please complete the following information):**
N/A

OzzieIsaacs:
I don't see the problem with the normal download link. The user using kobo sync should of course have download rights and for me the sync is working fine. If it's not working, this looks to me like it's a problem with yunohost.

OzzieIsaacs:
Your PR is merged, I consider this as solved

PR #1694, opened 2020-10-31
When serving a download URL via KoboSync, use a URL that is specific to Kobo

naydichev:
Related to #1693 - using a kobo specific download link.

I just looked at the code and haven't tested it yet, but I can work on that if this is has any likely hood of being merged.


OzzieIsaacs:
I don't see a reason for changing it.

naydichev:
I have very strict controls on access to my server - I'd rather only open the "kobo" URL to the world, rather than all of calibre. It seems reasonable to me to have all Kobo related things go through kobo specific endpoints.

PR #1704, opened 2020-11-09
Update jszip to 3.5.0. Update epub.js to 0.3.88.

GruberMarkus:
Fixes several problems when viewing EPUBs in a browser, for example when jumping to a certain chapter in the table of contents does not work.


OzzieIsaacs:
Thank you

PR #1720, opened 2020-11-21
Adapt find version path to actual repository structure

Scriptkiddi:
First of thanks for you work :D
I'm trying to package calibre web for nixos and found some issues with the setup.py.
The changes contain my suggestions for how to fix them, but I'm not really familiar with the pypi packaging process. Let me know what you think

- the setup.py expects a file structure that is not present in this repo. I think it should only use folders that are available.
- I also seen that your versioning schema does not conform to PEP 440 versioning https://www.python.org/dev/peps/pep-0440/
- The `console_script` is used for Python Functions not for Python scripts https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

### DIFF
--- a/cps/constants.py
+++ b/cps/constants.py
@@ -129,7 +129,7 @@ def selected_roles(dictionary):
 BookMeta = namedtuple('BookMeta', 'file_path, extension, title, author, cover, description, tags, series, '
                                   'series_id, languages')
 
-STABLE_VERSION = {'version': '0.6.10 Beta'}
+STABLE_VERSION = {'version': '0.6.10b1'}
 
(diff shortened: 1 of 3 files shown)


OzzieIsaacs:
I'm having a build script for packaging calibre-web (https://github.com/OzzieIsaacs/calibre-web-test/blob/master/build/make_release.py) which fixes this problem. My goal is to provide the script as is for direct download and installation and another packaged version which has located it's config in the user's home directory and still is able to update itself, I'm almost finished whit it (the detection of both modes isn't implemented and working). So I think there is no need for this PR

PR #1722, opened 2020-11-22
Forward Overdrive requests to Kobo Store

jathak:
When loading books from OverDrive, Kobo readers make requests to `/v1/products/books/external/<overdrive ids>`.
This adds a route to forward these requests to the Kobo Store, so that OverDrive still works when using Calibre-Web's Kobo sync server.

I've tested this manually with my Libra H2O connecting to my Calibre-Web server running on Ubuntu 20.04.1. Without this patch, most OverDrive pages would not load and new loans would not be downloaded during a sync, but after adding this everything seems to work as expected.


labor4:
It seems I cannot get any results with Overdrive.

Can you please make an example what should work?
What is the expected behaviour?

Me:
- can connect to Kobo store (being proxied)
- cannot load a overdrive page (settings neither, but login worked initially)
- syncing won't list/download loans made via browser GUI
- search in overdrive fails (book is available in overdrive browser gui)

Issue #778, opened 2019-01-18
caliblur for epub reader


leram84:
We need the caliblur.js/css loaded on the epub reader. We have some code in there to add an `X` button to the top right so that a user can close out the reader and revert back to the book details page, but its not loading currently. 

There might be a few adjustments we need to make once its in place to adapt it from the subfliter method we were using before, but first we need it loaded.

Thanks!

gelsas:
Does caliblur mean also a dark version of the reader in caliblur colors?

leram84:
no, we might do it for cw itself, but it doesn't make sense to do that in caliblur

OzzieIsaacs:
Calibre-web has 4 readers:
- pdf Reader
- epub Reader
- txt Reader
- comic Reader (for cbr, cbz, cbt files)
The comic reader already has a dark mode:
![grafik](https://user-images.githubusercontent.com/6438895/51424931-5e902600-1bd5-11e9-88af-b7092776e260.png)
![grafik](https://user-images.githubusercontent.com/6438895/51424935-6d76d880-1bd5-11e9-9e15-bd96f6b7ec20.png)
I think it would be great to have something similar (also with the user handling) for the epub reader, so there is no break in the design.

adocampo:
I'm agree. It's not important, but would be nice to have an X to close the reader.
I found it weird the first time when I didn't find the cross to close the reader, and I was afraid that closing the tab, would close CW itself. Once closed the tab, I realized it had opened a tab for the reader, which I didn't noticed at first. I guess a simple cross can help to avoid those situtations.

And as you all know, I up vote for a dark reader and let set it on CW, as per user and/or globally (ideally both).

hexeth:
I would start by updating epub.js

The current versions of epub.js appears to support themes, but the version used in CW is too old. There's an example doc [here](https://github.com/futurepress/epub.js/blob/master/examples/themes.html) as well as the documentation [here](http://epubjs.org/documentation/0.3/#themes).

It seems like if I'm reading the above correctly, themes can be registered and triggered via custom buttons.

OzzieIsaacs:
Juhu @adx442, 
here is the feature you requested

Jakgab617:
I was looking at the dark mode possibility for the epub reader.  I tried changing colors in main.css and popup.css (reader window) just to see what will happen. The menu, panel etc changed. But the actual book background and text color remained the same even with color declaration and !important tag.

[...]

PR #1736, opened 2020-12-01
Added basic dark mode support to the ePub reader

NickyBoy89:
I added a toggle button that allows toggling of dark mode in the ePub reader through the built-in settings menu

However, the only drawback that I had was the `iframe` that contains the content for the ePub gets refreshed after every section, which wipes the temporary css changes that changed its colors. The best way I found to solve this is to check the background of the `iframe` every time a page is turned

Suggestions welcome!

### DIFF
--- a/cps/static/js/reading/epub.js
+++ b/cps/static/js/reading/epub.js
@@ -46,17 +46,29 @@ var reader;
     // Dark mode logic
     $("#colorMode").on( "click", function() {
       if ($("#colorMode").prop("checked")) {
-        $("#main").css("background", "#343233");
-        $("#titlebar").css("color", "#ccc")
-        $("iframe").contents().find("body").css('background-color', '#4b4b4b');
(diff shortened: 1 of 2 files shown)


OzzieIsaacs:
Did you have a look at this one?: https://github.com/janeczku/calibre-web/issues/778#issuecomment-456560756

NickyBoy89:
@OzzieIsaacs Oh shoot, didn't see that one. It didn't come up in my preliminary search, but looking through it, there are a lot of good resources to use, and it seems that I should also look through what they did in the comic viewer to get dark mode support, which I assume uses a similar method?

OzzieIsaacs:
I didn't do the reader for epub and the dark mode for the comic reader also wasn't implemented by me, so I don't know. It's possible that the comic reader uses something totally different, as it's based on a different base (kthoom  vs. futurepress.epub.js)

Issue #904, opened 2019-05-03
CaliBlur theme - Custom columns not visible


a-maggioni:
When I use I the standard theme I can see all the custom columns I created in Calibre, while the same does not hold for caliblur theme.
![standard](https://user-images.githubusercontent.com/38586718/57176725-ab81ca80-6e5b-11e9-8dcd-7665921a469d.png)
![caliblur](https://user-images.githubusercontent.com/38586718/57176726-b177ab80-6e5b-11e9-8225-7eaef35bdd8e.png)

OzzieIsaacs:
I created a commit to make the custom columns visible.
They are not really good aligned right now:
![grafik](https://user-images.githubusercontent.com/6438895/58619232-c6155b00-82c4-11e9-9e9d-5d80aa4afa6f.png)
@leram84 , @hexeth could you please take a look at this issue. In the latest commit I renamed the custom_columns to real_custom_columns so it can be found easily. By doing that I realized, that by changing the language the texts are not really good aligned (Sprache(Language) and Herausgeber(Publisher) in the screenshot)

OzzieIsaacs:
Should work now

PR #1759, opened 2020-12-16
Preserve publisher links, and fix formatting in custom columns for CaliBlur theme

mmonkey:
On the CaliBlur book details page:

* resolves formatting issues described in #904 (custom columns should match the other columns)
* preserves publisher link (can now click on publisher to view books by that publisher)

Capture


OzzieIsaacs:
Thank you very much and sorry for the delay

PR #1768, opened 2020-12-23
Added ability to not truncate title

### LINKED ISSUE #1482: Book titles are truncated in details view
When going into Book Details, the title is truncated to ~60 characters. This means that a title such as "Imperial Twilight - The Opium War and the End of China’s Last Golden Age" becomes "Imperial Twilight - The Opium War and the End of China’s Last ...". 

It really should show the full title, though. Unlike the list/cover view, there's no reason to restrict the length.

nichwall:
Fixes #1482 

This PR adds an optional parameter to the `shortentitle_filter` function for Jinja to prevent shortening titles in the Details page. I tested it with both themes, and it looks like book titles are still shortened correctly everywhere else. The caliBlur theme (second image) shows the title is not shortened in the main view, but is still shortened just below the search bar.

Title not truncated:
image
image

Title still truncated in main view:
image

I did find an issue when browsing with a mobile device where the 26 characters of the english alphabet were too long to fit on the screen (shown below), but that is a different issue from the title being shortened.
image

### DIFF
--- a/cps/jinjia.py
+++ b/cps/jinjia.py
@@ -51,12 +51,14 @@ def url_for_other_page(page):
 
 # shortentitles to at longest nchar, shorten longer words if necessary
 @jinjia.app_template_filter('shortentitle')
-def shortentitle_filter(s, nchar=20):
+def shortentitle_filter(s, nchar=20, truncate_title=True):
     text = s.split()
     res = ""  # result
     suml = 0  # overall length
     for line in text:
-        if suml >= 60:
+        # truncate title if more than 60 characters and we don't want to show
+        # the entire title
+        if suml >= 60 and truncate_title:
             res += '...'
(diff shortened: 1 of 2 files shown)


OzzieIsaacs:
I think if we want to remove thre truncating, we just change: 
`<h2 id="title">{{entry.title|shortentitle(40)}}</h2>`
to
`<h2 id="title">{{entry.title}}</h2>`

nichwall:
That's fair. I originally made that change to fix the issue. After thinking about it a bit more, I thought it would be better to modify the function in case a case came up where the function should still be called without actually truncating the title. But, I think your comment would be the better change if there's not already a specific use case.
Do you want me to change this PR to do that, or just close the PR and let you do the single line of change?

PR #1796, opened 2021-01-04
Added proper UI support for dark mode on the ePub reader

NickyBoy89:
This is an implementation of UI support for dark mode in the ePub reader, that only includes the HTML and CSS changes of the settings menu

This continues the discussion in my previous PR #1736

Although, to properly add a theme for `epub.min.js`, the configuration code needs to be put in a JS file, because right now it appears that these functions are part of a minified JS file at `cps/static/js/libs/reader.min.js`, which I do not know how is generated

### DIFF
--- a/cps/static/css/kthoom.css
+++ b/cps/static/css/kthoom.css
@@ -181,7 +181,6 @@ th {
 }
 
 .settings-column {
-  float: left;
   min-width: 35%;
   padding-bottom: 10px;
 }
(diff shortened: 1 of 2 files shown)


NickyBoy89:
@OzzieIsaacs Do you know if the reader.min.js file is generated anywhere in the code?

OzzieIsaacs:
Should be this one: https://github.com/futurepress/epubjs-reader/blob/master/reader/js/reader.min.js. I never did change anything in the reader

OzzieIsaacs:
This guy there seems to have themes applied to the reader:
https://github.com/johnslion/epubjs-reader/commit/4e7ef5e149b471a59592a27f74787ad765081f54

OzzieIsaacs:
This PR only adds the menus but no functionality is added. Please make the functionality working, otherwise I can't merge the PR and I'll close it.
The kthoom.css is for and from the comic reader please do not mix it with the epub reader (kthoom is a lib as epubjs is also. I don't want to have relationships/dependencies between epub reader and comic reader functions)

Issue #1276, opened 2020-03-29
Kobo sync and large libraries


shermp:
This is both a problem report, and a feature request(s) combined, hence the blank issue.

*The Problem*
My Calibre library now has a few thousand books in it (it's been going for 12 years now...), and my Aura H2O simply times out when attempting to sync with calibre-web. The Kobo eventually gave a 'sync failed' message, and my wireshark capture shows that it gave up at around ~286K characters.

There needs to be some method of making large libraries more usable for kobo sync. A few ideas off the top of my head:
- I notice the initial sync sends the full metadata for each book. Is this required? Or is the Kobo smart enough to request additional metadata later?
- How about "archiving" the library by default, and "unarchiving" the books we want to sync? I actually don't want thousands of books on my Kobo at a time...
- Is it possible to recreate the store API, such that one could browse one's Calibre library as though we were browsing the Kobo store, then "purchase" books we want to sync?

@shavitmichael do you have any ideas or thoughts on this?

shavitmichael:
You raise an interesting problem.

> I notice the initial sync sends the full metadata for each book. Is this required? Or is the Kobo smart enough to request additional metadata later?

It turns out the Kobo device already has a solution for this problem, we just haven't been making use of it yet. First some background:
We respond to the sync request with one (possibly) massive reply for all new books, tags, and state **since the last request**. We recognize which books are new to the device thanks to timestamps we pass around in the SyncToken. Each iteration of the protocol can be though of as updating the device to the timestamp of the last modified book, say _t_latest_.
If necessary, we can instead update the device to some earlier timestamp _t_earlier_; so that the device catches up to _t_latest_ over multiple rounds instead of the single massive one  we currently have. It turns out that setting the "x-kobo-sync" header to "continue" informs  the device that it's not fully up to date, and causes it to queue another call to the sync endpoint.

---------
> How about "archiving" the library by default, and "unarchiving" the books we want to sync? I actually don't want thousands of books on my Kobo at a time...

That's definitely doable. I only added "Archiving" to CalibreWeb to support library management for the Kobo, and given your use-case it might have been slightly ill-conceived.  Perhaps "Kobo Library View" might be a better name?
We could add a user-controllable toggle so that it can either control:
1. Books on the device (for you :-) ).
2. Books not on the device (for me :-) ).

---------
> Is it possible to recreate the store API, such that one could browse one's Calibre library as though we were browsing the Kobo store, then "purchase" books we want to sync?

[...]

PR #1852, opened 2021-02-06
Added option in order to synchronize only selected shelves on Kobo device

alfred82santa:
* Added a flag to shelves in order to set whether they must be synchronized with Kobo device.
* Only books on kobo shelves are synchronized.
* Shelves modified after last sync with no kobo sync flag are remove from Kobo device.
* Update db on startup: kobo_sync field to shelf table.
* In order to keep old behavior, if no shelves are marked to kobo sync, all books and shelves are synchronized.

Maybe, it solves/helps: #1276

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -171,7 +171,7 @@ def HandleSyncRequest():
                             ub.BookShelf.date_added > sync_token.books_last_modified))
                 .filter(db.Data.format.in_(KOBO_FORMATS))
                 .order_by(db.Books.id)
-                .order_by('last_modified')
+                .order_by(ub.ArchivedBook.last_modified)
                 .join(ub.BookShelf, db.Books.id == ub.BookShelf.book_id)
                 .join(ub.Shelf)
                 .filter(ub.Shelf.kobo_sync)
@@ -253,11 +253,36 @@ def HandleSyncRequest():
         books_last_id = -1
 
     # generate reading state data
+    changed_reading_states = ub.session.query(ub.KoboReadingState)
+
+    if only_kobo_shelves:
+        changed_reading_states = (
+            changed_reading_states.join(ub.BookShelf, ub.KoboReadingState.book_id == ub.BookShelf.book_id)
+            .join(ub.Shelf)
+            .filter(
+                ub.Shelf.kobo_sync,
+                or_(
+                    func.datetime(ub.KoboReadingState.last_modified) > sync_token.reading_state_last_modified,
+                    ub.BookShelf.date_added > sync_token.books_last_modified
+                )
+            )
+        ).distinct()
+
+    else:
+        changed_reading_states = (
+            changed_reading_states.filter(
+                func.datetime(ub.KoboReadingState.last_modified) > sync_token.reading_state_last_modified
+            )
+        )
     changed_reading_states = (
-        ub.session.query(ub.KoboReadingState)
-        .filter(and_(func.datetime(ub.KoboReadingState.last_modified) > sync_token.reading_state_last_modified,
-                     ub.KoboReadingState.user_id == current_user.id,
-                     ub.KoboReadingState.book_id.notin_(reading_states_in_new_entitlements))))
+        changed_reading_states.filter(
+            and_(
+                ub.KoboReadingState.user_id == current_user.id,
+                ub.KoboReadingState.book_id.notin_(reading_states_in_new_entitlements)
+            )
+        )
+    )
+
     for kobo_reading_state in changed_reading_states.all():
         book = calibre_db.session.query(db.Books).filter(db.Books.id == kobo_reading_state.book_id).one_or_none()
         if book:
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -165,14 +165,14 @@ def HandleSyncRequest():
     if sync_token.books_last_id > -1:
         changed_entries = changed_entries.filter(db.Books.id > sync_token.books_last_id)
 
-    only_kobo_shelfs = (
+    only_kobo_shelves = (
                            calibre_db.session.query(ub.Shelf)
                                .filter(ub.Shelf.user_id == current_user.id)
                                .filter(ub.Shelf.kobo_sync)
                                .count()
                        ) > 0
 
-    if only_kobo_shelfs:
+    if only_kobo_shelves:
         changed_entries = (
             changed_entries.join(ub.BookShelf, db.Books.id == ub.BookShelf.book_id)
                 .join(ub.Shelf)
@@ -244,7 +244,7 @@ def HandleSyncRequest():
             })
             new_reading_state_last_modified = max(new_reading_state_last_modified, kobo_reading_state.last_modified)
 
-    sync_shelves(sync_token, sync_results, only_kobo_shelfs=only_kobo_shelfs)
+    sync_shelves(sync_token, sync_results, only_kobo_shelves=only_kobo_shelves)
 
     sync_token.books_last_created = new_books_last_created
     sync_token.books_last_modified = new_books_last_modified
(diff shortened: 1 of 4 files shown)


robjloranger:
I was just thinking how this would be a perfect solution to multi user Kobo sync

pulitux:
I like this aproach, almost perfect solution to large libraries sync

alfred82santa:
@OzzieIsaacs could you check it?

pulitux:
I've not so much spare time, but i'll try. i'm running a docker container but I'll try to make an image

rolfberkenbosch:
I have tested your code, nice work!!!

I found a small flaw, when you have removed books from the shelves and from your kobo. The cps keep syncing some information about the removed books from the shelves. That kind of information will eventually overwhelming the the sync duration/time and gives timeouts. I have tested that with alot of books and can confirm that. For each book that have been deleted from the self you will see the following information in the sync output:

```json
   {
      "ChangedReadingState":{
         "ReadingState":{
            "Created":"2021-01-18T20:03:46Z",
            "CurrentBookmark":{
               "LastModified":"2021-03-12T21:01:19Z"
            },
            "EntitlementId":"67b43ab5-ebad-4dae-8aed-ffa3ac26a3f8",
            "LastModified":"2021-03-12T21:01:19Z",
            "PriorityTimestamp":"2021-03-12T21:01:19Z",
            "Statistics":{
               "LastModified":"2021-03-12T21:01:19Z"
            },
            "StatusInfo":{
               "LastModified":"2021-03-12T21:01:19Z",
               "Status":"ReadyToRead",
               "TimesStartedReading":0
            }
         }
      }
   },
```

Can you remove that kind of information so that the sync keeps clean and fast only for the books that are on the shelves.

alfred82santa:
Could you try again, please? I think I've solved it.

mlaccetti:
Any chance this is going to get merged soon? Super excited to finally be able to sync to my Kobo!

OzzieIsaacs:
I have to finish some test on the development branch and then I’m ready for this one

rolfberkenbosch:
I have tested this version and looks ok for me

OzzieIsaacs:
I started working on it. My findings so far:

- [ ] you removed `new_tags_last_modified = max(shelf.last_modified, new_tags_last_modified)` from line 663 which causes deleted Tags to be synced every time unless you do another shelf operation
- [ ] I replaced the join in line 690 `for shelf in ub.session.query(ub.Shelf).join(ub.BookShelf).filter(` with an outer join, otherwise a new empty shelf isn't synced.
There is nothing to do for you right now, I fiexed both and will now start testing the new functionality

alfred82santa:
Well, about the first issue. I decided to remove all shelves because there is no way to know which shelves had have the flag in previous sync, and not in current sync. I thought it was small price to allow to remove synchronised shelves.
About second issues. Do you think we need to sync empty shelves?

OzzieIsaacs:
Second issue: it‘s a change to the previous behavior, I‘m running automated tests for it, I think I checked it with the real sync for how it should work

OzzieIsaacs:
Regarding first issue, I think we might misunderstood each other. My use case is the following: 
I'm having a shelf and no shelf is ticked "synchronize to Kobo", in this "mode" all shelfs and all books are getting synchronized. During sync this shelf will get transferred to Kobo (correct). Now I delete the shelf and sync again. The shelf will be synced as "deleted" shelf (correct). If I now do nothing but another sync, nothing should be synced (behaviour before) and from my point of view the right way. Because you deleted the above line kobo doesn't update the sync time for shelfs and the deleted shelf is again synced. Be putting back the line into the code, the shelf will be reported as deleted during one sync and nothing more. This has nothing to do with your changes, it's just the old functionality.

OzzieIsaacs:
One more thing:
Before the code change we had one use case:
All books and shelfs are synced to Kobo.

Now we are having 3 use cases:
All books are synced to Kobo, because there is no shelf
All books are synced to Kobo, because no shelf has a set the option `Synchronize with Kobo device`
Only books in shelfs with option `Synchronize with Kobo device` are synced ot Kobo.

If no Shelf has set the option `Synchronize with Kobo device` the user sees the following screen during editing of Shelf propertiers:
grafik
But the message is wrong, because this shelf is going to be synchronized, as long there is no other shelf having the `Synchronize with Kobo device` option set. 
I think this could confuse users. There is no place where you can see which of the above use cases you are currently in, you have to go through every Shelf to find out what's going to be synced.

What do you think?
To avoid this I split the message into 3, depending on the current status:
`Sync with Kobo device - If this option is turned off, all books and shelves will be synced with Kobo` -> displayed if currently displayed shelf it the only one to be synced
`Only sync books from selected shelves to Kobo instead of all books and shelves` -> displayed if no shelf has the kobo_sync option enabled
`Sync with Kobo device` -> in all other cases (more than one shelf is going to be synced, or one, but not the currently displayed shelf is synced

alfred82santa:
At this point, maybe we should change the main flow. I mean, we should add a new flag in general config in order to choose if user wants to sync all books and shelves or just selected ones. If flag is false the behaviour is the old one and the flag on shelves must disappear. If flag is true, the flag on shelves appears and just shelves with this flag set will be sync. 

Just changing this line:  https://github.com/janeczku/calibre-web/pull/1852/files#diff-07a86a0f46bee91e3fe947bec8a613581aaa97fb5d1430199bc784844701c10cR156

```python
only_kobo_shelves = config.config_kobo_shelves_sync # New config flag
```

edillmann:
Hello, any news ? This seem's perfect for my usage.

Cloudworks99:
Am I correct in assuming that this change is now available in v0.6.12 Pilar?

If so I tried testing it through the linuxserver.io docker build. I can switch the Kobo sync on under settings and I can generate a token for a user but I do not see the ""Synchronize with Kobo device" switch on the Shelf edit screen. I only see "Title" and "Share with Everyone". Any idea what I am missing to make this work?

mlaccetti:
If I remember correctly, I had to first create a new shelf and enable the "sync to Kobo" flag on the shelf. I could then add books to it and have them sync.

quamok:
> Am I correct in assuming that this change is now available in v0.6.12 Pilar?
> 
> If so I tried testing it through the linuxserver.io docker build. I can switch the Kobo sync on under settings and I can generate a token for a user but I do not see the "Synchronize with Kobo device" switch on the Shelf edit screen. I only see "Title" and "Share with Everyone". Any idea what I am missing to make this work?

Were you able to make it work? I'm having the same issue, running v0.6.12. Maybe it's just available in the nightly builds for now? Are they stable enough for daily usage?

Cloudworks99:
> Were you able to make it work? I'm having the same issue, running v0.6.12. Maybe it's just available in the nightly builds for now? Are they stable enough for daily usage?

I was not able to get it to work with the "latest" version of the linuxserver.io image.  But using the nightly version of that image I got it working. The code for it is in master but not in the v0.6.12 release.

quamok:
> > Were you able to make it work? I'm having the same issue, running v0.6.12. Maybe it's just available in the nightly builds for now? Are they stable enough for daily usage?
> 
> I was not able to get it to work with the "latest" version of the linuxserver.io image. But using the nightly version of that image I got it working. The code for it is in master but not in the v0.6.12 release.

Weird, I switched to nightly but still can't see the option...

OzzieIsaacs:
Please enable the feature first

grafik

quamok:
> Please enable the feature first
> 
> grafik

Thank you, that is obviously the answer. 

Is it supposed to delete books from the Kobo when a book is removed from the shelf? It seems to sync new books from the shelf to the Kobo device, but not removing books from Kobo that were removed from the shelf.

Stvn21:
> > Please enable the feature first
> > grafik
> 
> Thank you, that is obviously the answer.
> 
> Is it supposed to delete books from the Kobo when a book is removed from the shelf? It seems to sync new books from the shelf to the Kobo device, but not removing books from Kobo that were removed from the shelf.

I encounter the same -> after removing books from the shelf it doesn't get removed from the kobo device. not sure if this is expected behavior.

Issue #1691, opened 2020-10-30
Support for custom identifiers


Zaroz:
On a similar vein to [issue #550](https://github.com/janeczku/calibre-web/pull/550), I would like to request that support for custom identifiers be added. As it is, only a select few identifiers are supported (isbn, goodreads, url,), which limits a user's ability to link to the book's various metadata sites.

Currently, if you try to add a custom identifier, the link just links back to the book page, even if you put in a full url.

I am not asking for the whole system to be immediately added with rules, parsing, etc, just the ability to add custom ids with urls and have them show up as properly working hyperlinks by using the info added to the forms like (if it were written in html):
`<a href="{Identifier Value}">{Identifier Type}</a>`

Zaroz:
Since to apparent changes had been made, I decided to make a pull request myself. see: #1877 
Please approve it so I

Zaroz:
added changes via #1877

Zaroz:
awaiting approval of pull request of #1877

gorgobacka:
It seems that #1877 is approved.

PR #1877, opened 2021-03-03
modified to support custom url identifiers

Zaroz:
changed __repr__ else statement to resolve custom url entries

This allows adding multiple of any kind of url links that you want.


Zaroz:
created to resolve #1691

OzzieIsaacs (changes requested):
Now it's the same return value as for format_type == "url", please remove the elif format_type == "url", to make it one statement

Issue #1863, opened 2021-02-15
Mass add-to-shelf button not working in dark theme


hikanwoi:
**Describe the bug/problem**
When using the dark theme, in the advanced search result page, clicking the "add-to-shelf" button does not bring up the "add-to-shelves" dropdown menu. It works in the standard theme.

**To Reproduce**
0. Use dark theme
1. Go to advanced search
2. Do any search
3. Click the "add-to-shelf"
4. The dropdown menu does not show up as expected.

**Expected behavior**
The "add-to-shelves" dropdown menu should show up, allowing mass add to shelf.

**Screenshots**
http://prntscr.com/zl8txe

**Environment (please complete the following information):**
 - OS: Windows 10
 - Python version: 3.7.3
 - Calibre-Web version: 0.6.11
 - Docker container: None
 - Special Hardware: n/a
 - Browser: Firefox 85.0.2, Chrome 88

OzzieIsaacs:
The dark theme has indeed some issues. I'm not the programmer of it, so it's always a little bit behind the standard theme

cbartondock:
It actually does work in that if you go in to inspect the page and click the link it works fine. The problem is that the dropdown is getting hit with a "display: none" style for some reason.

jamestomasino:
I'm still experiencing this issue, even on nightly builds.

PR #1885, opened 2021-03-07
Improved Advanced Search

cbartondock:
* Added functionality to include / exclude books belonging to specific bookshelves
* Added select/deselect all buttons in dropdowns (excluding for tags because selecting all tags breaks things?)
* Fixed the fact that dropdowns would sometimes drop "up", potentially hiding the search box.
* Fixed [issue 1863](https://github.com/janeczku/calibre-web/issues/1863)


cbartondock:
I made the changes necessary to integrate my work into the recent refactor of `web.py`.

PR #1927, opened 2021-04-01
Added Google Scholar metadata capabilities

cbartondock:
- Google Scholar metadata: Extremely useful for academic papers and lectures. Requires the scholarly python library to be installed (`pip3 install scholarly`).
- Fixed the following graphical error in Caliblur (particularly bad on mobile). Also fixed the same graphical error for the case where more than one filetype is available to read in browser.
Capture
Now looks like
Capture2
- Removed the ugly outline that appeared on the left side of books when hovering over them in the main library and elsewhere. Could possibly be replaced with something better, but in the meantime it looked quite bad.
- Prevented dropdowns from getting cut off by a smaller webpage height, making it impossible to add things to your later shelves if you have a lot. They now just become scrollable when they're too long.


cbartondock:
Also fixed the graphical error for the case where more than one filetype is available to read in browser.

cbartondock:
Let me know if there's anything I should change. I thought maybe I should add scholarly to optional-requirementa.txt but I wasn't certain.

cbartondock:
Any comment on this PR? I have a few more features I would like to add but I am waiting until this is merged.

OzzieIsaacs:
I'm still struggling with the user mass edit feature. After this I will stark work on kobo sync pull request, than I have time to look at this one.

cbartondock:
Okay no worries ! Mass edit sounds ridiculously useful.

PR #2035, opened 2021-06-19
Update French translations

Thovi98:
The French translation is normally completely now.

I hope it’s all fine.

Thank you for this awesome app !!


OzzieIsaacs:
Please don't translate the texts with parenthesis like the following `%(error)s"`. These are format strings, error is a variable which is replaced later on with some text, if translated, the app will crash

OzzieIsaacs:
I fixed the problems. Thank you for your contribution.

PR #2099, opened 2021-08-26
Update messages.po

danielamazza:
Changed translate of "Read Books": the meaning is exactly "Libri letti", not "Libri da leggere"


OzzieIsaacs:
Thanks

PR #2109, opened 2021-09-08
feat: add mouse wheel support for epub reader

JeffersonQin:
I added a feature for epub reader by directly editing `reader.min.js`. It would enable us to use mouse wheel to flip pages. To explain what I did, I would show the complete code here. In the `EPUBJS.reader.ReaderController` section of the file, I wrote an event handler for scrolling functions, and made it support mainstream browsers. Here are the originial code:

```diff
...
p = function (b) {
	window.alert("here!");
	37 == b.keyCode && ("rtl" === a.package.metadata.direction ? h.next(): h.prev(),
		f.addClass("active"), (o = !0),
		setTimeout(function () {
			(o = !1), f.removeClass("active");
		}, 100),
		b.preventDefault()),
	39 == b.keyCode && ("rtl" === a.package.metadata.direction ? h.prev() : h.next(),
		e.addClass("active"), (o = !0),
		setTimeout(function () {
			(o = !1), e.removeClass("active");
		}, 100),
		b.preventDefault());
- };
+ },
+ eventHandle = {
+ 	getEvent: function(event){
+ 		return event || window.event;
+ 	},
+ 	addEvent: function(element, type, handler){
+ 		if(element.addEventListener){
+ 			element.addEventListener(type, handler, false);
+ 		}
+ 		else if(element.attachEvent){
+ 			element.attachEvent('on'+type, handler);
+ 		}else{
+ 			element['on'+type] = handler;
+ 		}
+ 	},
+ 	getWheelDelta: function(event){
+ 		return event.wheelDelta ? event.wheelDelta : (-event.detail) * 40;
+ 	}
+ },
+ mouseHandle = function(event) {
+ 	event = eventHandle.getEvent(event);
+ 	var delta = eventHandle.getWheelDelta(event);
+ 	if (delta < 0) "rtl" === a.package.metadata.direction ? h.prev() : h.next();
+ 	else "rtl" === a.package.metadata.direction ? h.next() : h.prev();
+ };
return (
	document.addEventListener("keydown", p, !1),
+ 	eventHandle.addEvent(document, 'mousewheel', mouseHandle),
+ 	eventHandle.addEvent(document, 'DOMMouseScroll', mouseHandle),
...
```

Here is the according piece of code after minification:

```javascript
eventHandle={getEvent:function(event){return event||window.event},addEvent:function(element,type,handler){if(element.addEventListener){element.addEventListener(type,handler,false)}else if(element.attachEvent){element.attachEvent('on'+type,handler)}else{element['on'+type]=handler}},getWheelDelta:function(event){return event.wheelDelta?event.wheelDelta:(-event.detail)*40}},mouseHandle=function(event){event=eventHandle.getEvent(event);var delta=eventHandle.getWheelDelta(event);if(delta<0)"rtl"===a.package.metadata.direction?h.prev():h.next();else"rtl"===a.package.metadata.direction?h.next():h.prev()};return(document.addEventListener("keydown",p,!1),eventHandle.addEvent(document,'mousewheel',mouseHandle),eventHandle.addEvent(document,'DOMMouseScroll',mouseHandle)
```

Note that this feature does not apply to the area of `iframe`, more work should be done if we want to apply this feature to the central area.


JeffersonQin:
Guys, any thoughts? 😶

OzzieIsaacs:
I'm not so happy about modifying the reader code, as this has to be done after each new release of reader.js.
(I'm having a few things like this for bugs in the corresponding js files which are not fixed yet, but I like to minimize this)

JeffersonQin:
@OzzieIsaacs Got that, thanks for your reply

OzzieIsaacs:
Easiest solution is just change reading/epub.js and add 2 events like:
```
    $(window).on( "mousewheel", mouseWheelEvent);
    $(window).on( "DOMMouseScroll", mouseWheelEvent);

    // reader.on('mousewheel', mouseWheelEvent);

    // For Firefox
    // $("#main").on('DOMMouseScroll', mouseWheelEvent);

    function mouseWheelEvent(e) {
        reader.rendition.next(); //-> go to next page
        reader.rendition.prev(); //-> go to prev page
        console.log("wheel");
    }
```
The problem I see is (same also with your code) that the middle of the document isn't  working (the iframe), only the surrounding is working as scroll source.
But I'm willing to accept a PR with this "bug" not changing the epub.min.js file itself

PR #2112, opened 2021-09-08
Add Greek Book Languge to Greek Language

nikosbatsikas:


OzzieIsaacs:
It's complicated. 
The file you are changed is automatically generated from here: https://github.com/OzzieIsaacs/calibre-web-test/blob/main/build/translations/for_use_calibre_iso639_el.po
This file was not generated by me, but it's from calibre itself and is hosted on transifex:
grafik
So the highly offical way would be creating a transifex account for calibre and add the translation there and inform me, so I update my copy of the file, BUT: The transifex file has above 7000 language names listed and only 400 are used. (Sound like a job for this guy from greece named Σίσυφος)
Anyway, I changed my copy of the iso639 file for geek, the translation is now included in calibre-web

PR #2132, opened 2021-10-01
New filename sort for Comic Reader

GarckaMan:
Prior filename sorting for Comic Reader would not work correctly for filenames with numbers on it. Comic Reader would open a file like this:
- 1.png, 10.png, 11.png ... 19.png, 2.png, 20.png, 21.png ...

This new filename sorting would change it to:

- 1.png, 2.png ... 11.png ... 19.png, 20.png, 21.png ...

### DIFF
--- a/cps/static/js/kthoom.js
+++ b/cps/static/js/kthoom.js
@@ -180,13 +180,15 @@ function initProgressClick() {
 
 function loadFromArrayBuffer(ab) {
     var lastCompletion = 0;
+    const collator = new Intl.Collator('en', { numeric: true, sensitivity: 'base' });
     loadArchiveFormats(['rar', 'zip', 'tar'], function() {
         // Open the file as an archive
         archiveOpenFile(ab, function (archive) {
             if (archive) {
                 totalImages = archive.entries.length
                 console.info('Uncompressing ' + archive.archive_type + ' ...');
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Yeah, this was broken during the last update of the reader. Before the following search function was used:
```
function alphanumCase(a, b) {
    function chunkify(t) {
        var tz = new Array();
        var x = 0, y = -1, n = 0, i, j;

        while (i = (j = t.charAt(x++)).charCodeAt(0)) {
            var m = (i === 46 || (i >= 48 && i <= 57));
            // Compare has to be with != otherwise fails
            if (m != n) {
                tz[++y] = "";
                n = m;
            }
            tz[y] += j;
        }
        return tz;
    }

    var aa = chunkify(a.filename.toLowerCase());
    var bb = chunkify(b.filename.toLowerCase());

    for (var x = 0; aa[x] && bb[x]; x++) {
        if (aa[x] !== bb[x]) {
            var c = Number(aa[x]), d = Number(bb[x]);
            // Compare has to be with == otherwise fails
            if (c == aa[x] && d == bb[x]) {
                return c - d;
            } else {
                return (aa[x] > bb[x]) ? 1 : -1;
            }
        }
    }
    return aa.length - bb.length;
}
```
Let me doublecheck if this really does the same

PR #2147, opened 2021-10-24
Prevent virtual keyboard login auto capitalization

bernizt:
Stops virtual keyboards, such as the ones on mobile devices, to default to uppercase for the first letter of the login user input field.


OzzieIsaacs:
The login to calibre-web is not case sensitive, so the auto capitalization is only annoying. But anyhow it's merged

PR #2189, opened 2021-11-24
Feature: Long Strip Display for Comic Reader (kthoom.js)

GarckaMan:
### Features

**Long Strip Display**
- Toggle between **Single Page** Display or **Long Strip** Display
- 4px margin between each page
- **Long Strip** Display will update ToC and Progress Bar if a new page is scrolled into (either with left/right/up/down arrow keys or mouse wheel)
- Navigating using the ToC and Progress Bar will scroll you to that page (if **Long Strip** Display is selected)
- Changing from **Long Strip** to **Single Page** will scroll you to the top of current page you were reading

**Other technical changes**
- All pages will be loaded at the start regardless of display selection
  - Selecting **Single Page** will hide all images except the current image 
- Removed **Remember Position/Scroll to Top** option (I created the option to fix an issue that's no longer there)
- Toggling fullscreen will focus on the main container so you can use up/down keys immediately 
- Settings are loaded before the file so the theme can be applied immediately
- **Rotate/Flip/Scale/Direction** options all work correctly
  - **Rotate/Flip** options will redraw all images, so large files (>1GB) will take a few seconds

**Misc. information**
- This feature was inspired by MangaDex.org site
- Load time may be longer. I didn't experience it on my machine but it is possible
- This feature may need a bit of testing, specially files with images that can't be opened by the reader and really big files (>1GB)

**Screenshots**

image
---
image


OzzieIsaacs:
I found today time to check this PR (finally). I think there I found one case that's not handled:
If you have a reverse order manga (biggest image number is the first page) and switch to the right to left mode and then (or before) activate the long strip view, the comics are still showing in the old direction, meaning the first (biggest number) page is still the last one. My feeling tells me, that it should in this case be the first one. (This is alos not working for the thumbails and it wasn't working for the thumbnails before). 
Would it be possible for you to change this?

GarckaMan:
I was not able to repro your issue @OzzieIsaacs because I couldn't find the way to open a file in reverse order. Is there a Calibre Web option I missed?

OzzieIsaacs:
From my knowledge some manga comics are in reverse: Meaning the comic archive has the following files:
comic1.jpg -> last page
comic2.jpg -> page before last page
comicN.jpg -> first page

To display this comics correct the reader supports right to left view. In this case the left button shows the file with the next bigger number and the right button shows the file with the next lower number. In the long Strip mode the file with the lowest number is still displayed on top and if you want to read the comic in the right way you have to start from the bottom and scroll up.

My intention was to ask if it would be possible to reverse the visible order for this case?

GarckaMan:
Oh, I get it.
The thing is, that's a complete new feature outside of just the Long Strip functionality. The behaviour of Right to Left function is just to invert the left/right functions (left will advance the page, right will show previous one).

So, to answer your question: Yes it is possible, however, that would be another whole new functionality that (I think) doesn't belong in this PR.

PR #2204, opened 2021-12-01
Feature: Add metadata provider douban

acccccccb:
Add a new metadata provider in Fetch Metadata modal, it can get metadata from https://books.douban.com
It‘s very helpful to obtain Chinese book information

You can see it at here：
Edit Metadata -> Fetch Metadata ->  get metadata form Douban.com

----
在获取元数据弹窗中新增一个了一个选项，从豆瓣 https://books.douban.com 获取元数据
你可以在这里看到此功能：
编辑元数据 -> 获取元数据 -> 从豆瓣获取

### DIFF
--- a/cps/metadata_provider/douban.py
+++ b/cps/metadata_provider/douban.py
@@ -1,0 +1,195 @@
+# -*- coding: utf-8 -*-
+
+#  This file is part of the Calibre-Web (https://github.com/janeczku/calibre-web)
+#    Copyright (C) 2021 OzzieIsaacs
+#
+#  This program is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+# Douban Books api document: https://ihtmlcss.com
+# Code by tabzhang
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
I'm not sure about it. It takes a long time to generate results as it's doing a 2nd request for every entry it found.
As I understand it, it's relying on a cookie, which is tied to a user_id and some of the cookie content looks like timestamps. So my fear is that it will stop at a certain time and we have to change the api again. 
My idea with migrating everything to python and doing it more generic was that everybody could write it's own plugin and offer it somewhere and users just could download it (and if it stops working it's not my time). Currently I'm halfway through this process (it's currently not possible to keep "outside" files after updates)

acccccccb:
1. The first request can only obtain the book list, which contains only a small amount of information, so the second request must be executed to obtain detailed book information. I've limited the amount of content I can get to avoid making a lot of requests.

2. If the request does not contain cookies, the request will be rejected. According to my test, the server will not verify whether the cookies are invalid, but only whether there are cookies. So far, it works well on my server.

3. This API will be modified frequently in the future because its data comes from regular matching. If the content of the target website changes, I will have to modify the matching rules. If the cookie verification mechanism changes, maybe I must also modify cookies. These are uncertain.

4. It's a great idea to package it into a plug-in, but I don't know how to package it into a plug-in and use it in calibre web

----


1. 第一次请求只能获取到书籍列表，它只包含少量信息，所以必须执行第二次请求获取详细的书籍信息。我已对获取内容的数量做了限制以免发起大量请求。
2. 如果请求不带cookies，请求将会被拒绝，根据我的测试，服务器并不会验证cookies是否失效，只会验证是否存在cookies。目前为止，在我的服务器上，它运行良好。
3. 这个api在将来确实会有很频繁的修改，因为它的数据来源于正则匹配，如果目标网站的内容发生更改，我将不得不修改匹配規則，如果cookies验证机制有变化，或许我也必须修改cookies。这些都是不确定的。
4. 封装成插件是个很棒的想法，不过我不知道如何封装成插件并用于calibre-web

PR #2212, opened 2021-12-08
Remove "Delete selections" for checkbox in table

photonq2:
Button unnecessary. Checkbox in tables already there.

image
image

### DIFF
--- a/cps/static/js/table.js
+++ b/cps/static/js/table.js
@@ -64,21 +64,14 @@ $(function() {
                 $("#merge_books").attr("aria-disabled", true);
             }
             if (selections.length < 1) {
-                $("#delete_selection").addClass("disabled");
-                $("#delete_selection").attr("aria-disabled", true);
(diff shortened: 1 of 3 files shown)


OzzieIsaacs:
With this change it's one click more to deselect everything. The first click selects all elements on the page and the second one is deselecting everything. I agree that this is a more usual way of selecting things in a table.
I have to think on this

photonq2:
Keep in mind future bulk edit/delete buttons

webysther:
Remove 30 books some days ago...

OzzieIsaacs:
I meanwhile had time to look at the commit. Meanwhile with the current implementation it's possible to select books (books table) on different pages (at least until you reload the page). Your commit does break this.
In the background the javascript keeps a list of ticked books, to do it for all books an additional request for getting all "currently visible" books to tick them. I currently don't think the effort for making this possible doesn't match with the benefit for it, so I would close the PR. Sorry for the long waiting time

PR #2261, opened 2022-01-10
Update Korean translations

byword77:
Some mismatched translations corrected.
Fix all warning such as 'If the original text ends without a "." the translation normally also should end without it'

Thank you.


OzzieIsaacs:
Thank you, it's merged

Issue #1514, opened 2020-06-30
Blank margins in comics downloaded via kobosync


AntonioDeTeo:
Hi!,
This is a strange problem:
I have comics in cbr/cbz format and I've converted them to epub format so they can be downloaded via kobosync, but comics downloaded by this method show a white margin (except for the cover, which is shown at full screen), so the screen is not fully used.
Those same epubs, if I load them into my kobo using the windows Calibre application or directly copy them to the root of the device, they look fine without margins.
I don't know if it's possible to solve this from calibre-web or it's a kobo firmware problem that can't be fixed.
I have also tried to download them in kepub format and the result is the same.

I've tried it on a kobo libra H2O and a kobo forma.

Any idea to solve this? Does it happen to anyone else?

ATD555:
[Kobo uses the CBZ/CBR format for reading comics](https://help.kobo.com/hc/en-us/articles/360017763713-File-formats-your-Kobo-eReader-and-Kobo-Books-app-support). What happens if you keep it as CBZ/CBR format?

AntonioDeTeo:
Hello,
the point is that to download the comics using the kobosync functionality, the comics have to be in epub or kepub format. If they are not in some of those formats, they are ignored and not available for download by this method.
If I copy the comics in cbz/cbr/epub format directly via usb cable to my kobo device, I have no problem and they look good without white borders. 
The strange thing is that exactly the same comic (in epub format) if I download it using kobosync functionality, these margins appear and if I copy it directly from the pc to the device it looks without margins.

OzzieIsaacs:
Most likely related to #1219 and hard (if not impossible) to fix, as kobo device handles both ways of putting a book onto the device different.

manzaloros:
Just wanted to jump in and confirm that I have the same issue.

When loading a manga / comic directly onto the kobo with usb, there are no margins and the images are viewed full screen. When loading the same epub via kobo sync on Calibre Web, there are annoying white margins. At first I thought this was because kobosync was automatically converting the epubs to kepubs when sending over Kobo sync, and the kepub will always have white margins, but I’m not sure about that.

SpykerNZ:
I have encountered the same issue.

If I copy the epub directly to my kobo via usb then there are no margins.
If I use kobosync, then the white margins appear. During this sync the file is converted to a kepub.

As @manzaloros mentioned it might be the conversion to kepub causing this. Would there be a way to force sync as an epub?

vivaria:
[...]

PR #2262, opened 2022-01-10
Kobo sync detect fixed layout

### LINKED ISSUE #1219: [Kobo] Detect Fixed layout epubs
Kobo devices don't detect FixedLayout epubs when they are imported over the sync protocol.
CalibreWeb likely needs to parse epubs (and kepubs) in order to correctly set the epub's Format for the device.

Some research notes below
-----
### Why this happens:

When an epub/kepub is imported, it goes through the _EpubParser::ParseMetadata_ function, where an "EpubType" enum is set by parsing the .opf in the following fashion:
```
if (volume.epubType() == -1) {
  if (opf["rendition:layout"] == "pre-paginated") {
   volume.setEpubType(ENUM_EPUB3FL)
  }
  if (opf.has("omf:version")) {
    volume.setEpubType(ENUM_EPUB3_OMF);
  }
}
```
A book's EpubType subsequently controls whether it will be displayed as a [Fixed Layout Document](http://idpf.org/epub/fxl/).

When a book is imported over USB, the volume's epubType defaults to -1 and its epubType is therefore set using the above logic.
When a book is imported over the Sync protocol however, the volume's epubType has already been initialized based on the DownloadUrl's format string, in the following fashion:
```
switch ContentUrl.formatString():
  case "EPUB3FL":
    SetEpubType(ENUM_EPUB3FL)
  case "EPUB_KEPUB":
    SetEpubType(ENUM_KEPUB)
  case "EPUB3_OMF":
    SetEpubType(ENUM_EPUB3OMF)
  case etc...
  ...
   default:
      SetEpubType(-1);
```
### Can we trick the Kobo device into figuring out the EpubType for us?
In order to rely on the _EpubParser::parseMetadata_'s detection of FixedLayouts, we'd have to hit the default case in the format switch statement above. Unfortunately, the Kobo device will only consider a ContentUrl (a.k.a DonwloadUrl) for download by matching its format against one of the following format lists
```
List<String> format_list;
if content_type == CONTENT_TYPE_SAMPLE:
  format_list = ["EPUB3_SAMPLE", "EPUB3FL_SAMPLE", ...]
else if content_type == CONTENT_TYPE_FULL:
  format_LIst = ["EPUB3", "EPUB3FL", ...]
else if content_type == CONTENT_TYPE_INTERNET_ARCHIVE:
  format_list = ["EPUB"]
```
All of these have a matching case in the earlier switch statement except for "EPUB". And indeed we can import a book with an epubType of -1 by setting the "IsInternetArchive" field in the entitlement's BookMetadata json and setting the "format" field to "EPUB" in the downloadUrl.
This get's us really close but we're out of luck, it turns out the  _EpubParser::ParseMetadata_  function call in question is skipped over precisely when the IsInternetArchive is true :-( .

### Other hacks
In the meantime I've been using [kobopatch](https://github.com/geek1011/kobopatch) to removes the ` if (volume.epubType() == -1) ` check in the _EpubParser::ParseMetadata_ function.
The following patch worked on version 4.18.13737 for me:
```
Detect EpubType:
  - Enabled: yes
  - ReplaceBytes: {Offset: 0x796630, FindH: 01 30, ReplaceH: 01 28} # add R0, #0x1 -> cmp R0, #0x1
```
(To be clear this is *not* the recommended solution for this issue, just a stop gap I've been using for myself).

petersonev:
When using Kobo sync the layout of the Kepub has to be explicitly given, unlike when books are imported over USB which Kobo will check the epub format and set the type accordingly.

Currently the Kobo sync metadata format always returns `KEPUB` even when the epub has the layout set to `'pre-paginated'` for fixed layout epubs.

This PR checks if the layout is specified in the epub metadata and if the layout is set to be fixed will change the format accordingly to `EPUB3FL`.

This allows for epubs to take up the full screen which is useful for comics/manga and should fix #1219 and  #1514


DamiaPoquet:
@OzzieIsaacs just in case you missed it, this PR should fix some issues we are having with comic/manga files. It's interesting to get it merged.

PR #2288, opened 2022-01-25
added support for amazon metadata

quarz12:
added support for amazon metadata by using BeautifulSoup. Sadly it is quite a bit slower than just accessing an API (on an RPI4 a search takes about 10 seconds) as parsing the html can take a while. Also some metadata is only provided sometimes or not at all so for performance reasons I left that out completely. It works really well however if you only care about author and description. also of all metadata providers this should give the highest resolution cover.

### DIFF
--- a/cps/metadata_provider/amazon.py
+++ b/cps/metadata_provider/amazon.py
@@ -1,0 +1,98 @@
+import concurrent.futures
+import requests
+from bs4 import BeautifulSoup as BS  # requirement
+import lxml     #requirement for better speed
+import cchardet #optional for better speed
+from cps.services.Metadata import Metadata
+#from time import time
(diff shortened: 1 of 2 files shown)


cps/metadata_provider/amazon.py
```diff
@@ -0,0 +1,98 @@
+import concurrent.futures
+import requests
+from bs4 import BeautifulSoup as BS  # requirement
+import lxml     #requirement for better speed
+import cchardet #optional for better speed
```
OzzieIsaacs, line 5: It it's optional, please put a try: except ImportError statement around it, so users can use it without installing cchardet

quarz12:
done

PR #2420, opened 2022-05-22
Support usage of empty pre-created database

### LINKED ISSUE #2403: Creating calibre database via calibre-web
Calibre-Web is fantastic, and I even think for many users it has enough features to not need the main Calibre application anymore. That's why I think that for people with no pre-existing database, Calibre-Web should allow for the creatio of a new database via the UI. Having to install the main application to create a database feels like an artificial obstacle and might put off new users.

Allow for the creation of a new database via the UI.

Dunrar:
Calibre-Web only allows the usage of a database pre-created via Calibre. Some users may not be in need of all the advanced features of Calibre, and might not even have used it before. But they would have to install it nonetheless, just to create a database.

In order to circumvent this, this PR simply adds an empty, pre-created database to the source. When choosing a directory with no `metadata.db` file in it, the user is asked if he or she wants to "create" one. The database template is then simply copied over to the destination. Fixes #2403.

This implementation works, but is very basic. Some ideas that maybe should be discussed:
- Download the DB file instead of putting it in the source
- Add a `Create Database` button to the filechooser_modal -> Question: How/Where should error messages be displayed?
- Change implementation to use an `/ajax/db_create` route


OzzieIsaacs:
I think we should handle it easy and don't program some highly sophisticated functions into the code, but just have a link to a database file in the readme, so people can download it manually and put it wherever they want. 
This would be simple and easy to "program"

Dunrar:
Thanks for the feedback! I agree, simpler is better. But I also think there is a usability argument to be made for installing and using Calibre-Web without having to go through too much of the documentation. This PR doesn't add that much code and I think is a good tradeoff to make regarding simplicity of the code and ease of use. It's up to you, of course.

OzzieIsaacs:
I created a empty metadata.db from calibre and added it to the repository. Let's check if this helps.

PR #2421, opened 2022-05-22
Multiple authors and publication date in opds feed

### LINKED ISSUE #2402: Missing metadata in OPDS feed
When it comes to the OPDS feed, there are some differences to the one used by the calibre server. Information that seems to be missing (tested with Calibre Sync, FBReader, Librera and Moon+) includes:
- Multiple authors (co-authors)
- Publication date
- Identifiers (ISBN, ...)

Dunrar:
Adds some metadata to the opds feed (multiple authors and publication date). 

Some reader software picks up on this, others doesn't. What is being displayed and how differs greatly from reader to reader. For example, another way to display multiple authors is 
```
{% if entry.Books.authors.__len__() > 0 %}
    <author>
      <name>{% for author in entry.Books.authors %}{{author.name}}{% if not loop.last %} &amp; {% endif %}{% endfor %}</name>
    </author>
{% endif %}
```
While some can work with both others can only use one or the other. 

So while I'd really like to also add multiple identifiers (ISBN, DOI, Amazon, ...) I have not found a way to have any reader software pick up on it. I'm not sure if it is even worth implementing this. Would greatly appreciate input! Fixes #2402.

### DIFF
--- a/cps/templates/feed.xml
+++ b/cps/templates/feed.xml
@@ -53,7 +53,7 @@
         <name>{{entry.Books.publishers[0].name}}</name>
       </publisher>
     {% endif %}
-    <published>{{entry.Books.pubdate}}</published>
+    <published>{{entry.Books.pubdate.strftime("%Y-%m-%dT%H:%M:%S+00:00")}}</published>
     {% for lang in entry.Books.languages %}
       <dcterms:language>{{lang.lang_code}}</dcterms:language>
     {% endfor %}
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
I had time to check the PR:
The calibre contentserver outputs dates always in this format:
grafik
Where your dateformat looks like:
grafik


As some readers don't really follow the opds standard and seem only to check if it works with calibre I prefer to have the publisher date also formated like calibre does (this could be easily done with an addtional jinjia2 filter).

Could you please change it.
Thanks

Dunrar:
@OzzieIsaacs this should do it.

PR #2458, opened 2022-06-25
Fix Japanese translation

subdiox:
## Description
- Added 500+ Japanese translations
- Fixed 50+ Japanese translations

[*] I'm native Japanese.


OzzieIsaacs:
Thanks

PR #2463, opened 2022-06-29
Embed metadata on convert

Dunrar:
This is a draft PR that makes Calibre-Web embed metadata into files on conversion, just like Calibre does. Most of the code is refactoring how Calibre-Web deals with the Calibre binaries, which should also help with embedding metadata when sending or downloading books. For example, a directory containing all the supported Calibre binaries has to be chosen now, instead of the `ebook-convert` binary directly.

There are a few obvious reasons why this is a draft for now. First, the refactoring and actual embedding might have to go into separate PRs. Also, the code quality could definitely be improved, and I will do so when some more fundamental things have been discussed. 

Questions I have:
- Is the `get_calibre_binarypath()` function in `config_sql.py` okay, or should class variables be used for all supported/needed Calibre binaries?
- If we go with the `get_calibre_binarypath()` function, should `_ConfigSQL.config_converterpath` be kept or should `config.get_calibre_binarypath("ebook-convert")` be used everywhere?
- If we go with class variables, should each path be specified manually by the user, or is automatically setting them from a directory sufficient?
- Is including the `_extension` variable in `constants.py` okay?

Would be happy to discuss if anyone wants to chime in.


Dunrar:
So, I've cleaned up a little and decided to refactor `get_calibre_binarypath()` out of the `_ConfigSQL` class and into `helper.py`. The `_ConfigSQL.config_converterpath` DB column and variable still exists, but could be replaced by `helper.get_calibre_binarypath("ebook-convert")`. I will mark this ready fore review.

Dunrar:
@OzzieIsaacs Do you think you will be able to have a look at this in the near future? Thanks!

melyux:
@OzzieIsaacs This would be a great change and would enable #2497 to have the metadata edited from Calibre Web actually be applied to the books when reading them on devices

vampywiz17:
It would be great! If it happen, i delete calibre standalone forever!

Any ETA that it merge?

Appel-flappen:
Just tried this out on the nightly image and it's working for me. Many thanks for your work :)

vampywiz17:
Also working! Great job, thanks!

PR #2467, opened 2022-06-30
Add support for JWT based reverse proxy login headers

ruifung:
This enables reverse proxy login via CloudFlare Access.

Or well, in theory, any setup that provides a JWT that is signed by a public key obtainable via a JWKS URL.

Also allows for login header matching against username OR email if desired.

It should be noted that for now, I have not yet gotten around to adding translations for the additional setting items.


ruifung:
Closed because I messed up the git branches

PR #2479, opened 2022-07-08
Allow ereader_email to be CSV to allow sending to multiple E-Readers

gjutras:


valantur:
oh man I want this so badly!!

Issue #607, opened 2018-09-04
Feature Request: Polish/Embed metadata after applying metadata edits?


rmzg:
As the title says, if you have access to the calibre tools, use them to attempt to embed the metadata back into the actual ebook file.

This is something I was actually planning on working on but I wanted to know if there was some obvious reason this wouldn't work.

OzzieIsaacs:
In general there is no reason why this should not work

Dunrar:
I would like to try and implement this. @OzzieIsaacs would you prefer to use .opf files for this, like Calibre does, or would embedding the metadata directly suffice? Also, should this happen on download, metadata edits in general or via the GUI, whenever a user requests it?

Dunrar:
Okay, regarding A), I'll spend some time on finding out how Calibre deals with .opf files and metadata changes in general. Then we can more easily decide if we want to do it the same way or not. 
When it comes to B), I feel like any support is better than none, and people are complaining that there is no support right now. So how about starting with the most used or easiest ones to implement and incrementally add the others?

OzzieIsaacs:
Regarding A) excellent

> So how about starting with the most used or easiest ones to implement and incrementally add the others?

Yes, okay, in my opinion this would be pdf and epubs

OzzieIsaacs:
I read a bit in the calibre manual. New idea: 
Instead of starting the adventure to implement "applying metadata to an ebook by calibre-web", we should start with using calibre itself (https://manual.calibre-ebook.com/generated/en/calibredb.html#embed-metadata)

Dunrar:
I had some time to look into the Calibre source, so I'll just dump some info first:

On changes to book metadata, the update is made to the database first, and from there to the .opf files. When it comes to embedding metadata, it is as you said, every format has very different capabilities when it comes to storing metadata, and has to be handled seperately. 

Calibre has a few operations that embed metadata. Basically everything that moves files outside of Calibre and a few manual ones:
- Save/Export
- Send
- Polish
- Embed Metadata
- Conversion (sometimes, more on that later)

[...]

PR #2497, opened 2022-07-22
Embed metadata when sending to device

Dunrar:
This is a draft PR for embedding metadata when sending to devices. It depends on  #2463. At the moment the Calibre CLI is used to create OPF files, these parts of the code should be deleted once Calibre-Web supports OPF files.


OzzieIsaacs:
Also Issue: #607

Issue #2512, opened 2022-08-12
Kobo sync: Wrong ebook-language


beedaddy:
I'm very excited about the kobo sync feature and it works very well. But now I noticed that the device thinks, that all my (German) eBooks are in English. I double-checked that the language is correctly set in Calibre and so it is in calibre-web. But if I synchronize with my Kobo device, the Libra 2 shows "English" as the eBook's language. This is not a big problem, but the dictionary thus opens with the wrong translation mode EN -> DE. 

(using calibre-web 0.6.19)

PR #2513, opened 2022-08-12
kobo sync: using the actual book language

beedaddy:
The current approach is to simply set "en" as book language. This PR tries to use the actual language specified in the metadata of the book.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -45,6 +45,7 @@ import requests
 
 
 from . import config, logger, kobo_auth, db, calibre_db, helper, shelf as shelf_lib, ub, csrf, kobo_sync_status
+from . import isoLanguages
 from .constants import sqlalchemy_version2, COVER_THUMBNAIL_SMALL
 from .helper import get_download_link
 from .services import SyncToken as SyncToken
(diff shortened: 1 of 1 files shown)


beedaddy:
This should fix #2512

beedaddy:
I now realized, that the hyphenation does not work anymore for my German ebooks. The reason seams to be, that the language is set to **deu** in the content-table in KoboReader.sqlite instead of **de**…

OzzieIsaacs:
There is a function somewhere in language.py to convert 3 letter language code to 2-letter, so no Beinbruch

beedaddy:
Yes, I'm currently searching for such a function. I'm very new to the codebase, so … :)

OzzieIsaacs:
I thinks it's this one:
https://github.com/janeczku/calibre-web/blob/b5d5660d04acc98283b4ff4a798596d0d270441e/cps/isoLanguages.py#L42

import isoLanguages

lancode="deu"
result = isoLanguages.get(part3=langcode).part1

(I'm not on my computer so I can't test it)

beedaddy:
Thanks for the hint. Now it looks good here.

beedaddy:
@OzzieIsaacs Please let me know if there's anything else missing to get this PR accepted.

PR #2550, opened 2022-09-27
Fix bug of douban metadata_provider

ctaoist:
To fix the problem that the publisher and series are not right when using douban metadata provider.


xlivevil:
Still have a problem
Publisher maybe is a text
```html
<span class="pl">出版社:</span> 台灣角川<br>
```
or a link
```html
<span class="pl">出版社:</span>
      <a href="https://book.douban.com/press/2643">新星出版社</a>
    <br>
```
The old code only works in the first case, your fix ignores that.

PR #2559, opened 2022-10-02
update message.po (italian)

ElQuimm:
Hello, I ran the October 2 update, and this error appeared: 

Merge remote-tracking branch 'it/patch-28'
# Conflicts:
# cps/translations/it/LC_MESSAGES/messages.po

Calibre-Web does not start again and the container logfile does not allow me to have more information. I proceeded to verify the translation file (I did not find anything strange) and I made some updates of the text.

### DIFF
--- a/cps/translations/it/LC_MESSAGES/messages.po
+++ b/cps/translations/it/LC_MESSAGES/messages.po
@@ -38,9 +38,8 @@ msgid "Unknown command"
 msgstr "Comando sconosciuto"
 
 #: cps/admin.py:169
-#, fuzzy
 msgid "Books successfully queued for Metadata Backup"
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
This is not a error message from calibre-web (I don't know where it should show up). Looks like some git message. I modified the last translation file, because there was somebody else changing a English text. I think git detects your local change and it was not mergable with the remote branch and this caused the error. As I don't know what you are doing I don't know how to solve it.
One idea would be to search for patterns like the following and resolve the conflicts:
```
<<<<<<< HEAD
this is some content to mess with
content to append
=======
totally different content to merge later
>>>>>>> new_branch_to_merge_later
```

ElQuimm:
The update now work and Calibre-web run correctly.

Issue #2471, opened 2022-07-03
Task status not updated after succesfull sending email to kindle


therealactionadam:
**Describe the bug/problem**
After successfull sending Email to kindle task status not changed from started to finished.

**To Reproduce**
1. Choose Book (PDF)
2. Click "Send to Kindle"

**Screenshots**
![grafik](https://user-images.githubusercontent.com/21138161/177040442-57d517ab-3097-4c2e-8f30-d2a126d8677e.png)

**Environment (please complete the following information):**
 - docker 
 - gmail Oauth2 used

IHSten:
Also seeing this for MOBI's (even though Amazon is removing support for them soon)

ghighi3f:
I've noticed the problem myself and opened a PR with a fix.

OzzieIsaacs:
Please update to the newest nightly version, the problem is solved there (Linuxserver should provide a docker version for "latest")

PR #2560, opened 2022-10-04
Fix: issue 2471

ghighi3f:
Hello. 
I've noticed that when using the Gmail Account with OAuth2 email settings, 
the status of the task never changes from _Started_ when the email is sent successfully, for errors tho it changes correctly.
Same problem is mentioned here #2471.
Fix was pretty minor, I've just added the _handleSuccess() method call after the gmail function.
Lemme know if it's enough, in my tests everything works ok.
Thanks


OzzieIsaacs:
Thanks a lot

PR #2570, opened 2022-10-11
Comics save last position

OlivierMaire:
On each page change save the position as a bookmark.
When loading the comic, use the last known position.


Jerome2103:
@OzzieIsaacs when will this be merged? waiting for this features

PR #2576, opened 2022-10-19
Fix "Kobo eReader.conf" path

jeroenkroese:


OzzieIsaacs:
Thanks for pointing this out. But this is not the way it has to be changed. The html file in the template folder has to be changed and than the translation catalog has to be regenerated

jeroenkroese:
I used the Github IDE.. My bad. I see that it has been fixed now! I added a PR to also fix it for the translations: https://github.com/janeczku/calibre-web/pull/2638

PR #2580, opened 2022-10-21
fixes for 946, 2284 caliblur

### LINKED ISSUE #946: caliBlur! Register link not visible
Hey,

thanks for this amazing theme ! But when you setup open registration, then the register link is not visible at all.

Is there any way to fix this please ?

Thanks !

### LINKED ISSUE #2284: Unnecessary page refresh/blanking on Safari.
Ón safari, page loads are followed an unneeded extra blanking and redrawing of the page. It's adds a delay of about 2 seconds between seeing the new content and then waiting to interact (a bit frustrating, but I blame Safari).
cw2

By comparison, on Firefox the content just loads quickly without a fade effect and the user can begin interacting immediately.

Safari version: 15.2 on M1 Mac (native build).
Firefox version: 96.0.1 on M1 Mac (native build)

ghighi3f:
Resolves #946.
Most of the caliBlur ui is set using js after the html is loaded (really a hassle), so I tried to simplify the structure and logic because I couldn't fix it as it was.
For Login page I've added 2 buttons (register and home) for both possible settings (allow public registration, allow anonymus browsing)
Screenshot 2022-10-21 at 21 11 28
and for the register page and the rest of the site when user is anonymus, the links are grouped in the profile dropdown
Screenshot 2022-10-21 at 21 12 50

To simplify the logic and make it easier to customize/edit, I've added the _blur_ class on body in the html, so lot of css is applied immediately, and other small changes. 
The class also fixed #2284 (for me the refresh happended on chrome too)

I've tested the changes as much as I could do, hopefully everything is ok.

Let me know if anything
Cheers

### DIFF
--- a/cps/static/css/caliBlur.css
+++ b/cps/static/css/caliBlur.css
@@ -4423,38 +4423,6 @@ body.advanced_search > div.container-fluid > div.row-fluid > div.col-sm-10 > div
         left: 49px;
         margin-top: 5px
     }
-
-    body:not(.blur) > .navbar > .container-fluid > .navbar-header:after, body:not(.blur) > .navbar > .container-fluid > .navbar-header:before {
-        color: hsla(0, 0%, 100%, .7);
-        cursor: pointer;
-        display: block;
-        font-family: plex-icons-new, serif;
-        font-size: 20px;
-        font-stretch: 100%;
-        font-style: normal;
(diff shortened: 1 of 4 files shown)


OzzieIsaacs:
Thanks a lot

PR #2630, opened 2022-12-20
Add French articles to the title regex

jvoisin:


OzzieIsaacs:
Calibre generates (or generated) the regegex based on the language first used to start Calibre.For Calibre-Web we don't have this option. So this solution isn't ideal, but it's right german is not the only language in the world which needs the title regex

jvoisin:
I have a multilingual library, so autodetection based on my locale wouldn't fly anyway :)

PR #2636, opened 2022-12-25
Minor code refactorisation of epub.py

jvoisin:
- Reduce the amount of nested indentation
- Use proper functions instead of fragile manual parsing


OzzieIsaacs:
This: 
grafik
Lead to showing [] again for empty comments. I reverted it.
There was a typo:
grafik
should be "cs.endswith" instead of "s.endswith"
I prefer the code to return in all cases a value, so I also reverted this:
grafik
The other this are merged

PR #2646, opened 2023-01-06
Added norwegian translation

Allram:


OzzieIsaacs:
Thanks a lot. It's merged.

One remark, please don't translate the the string between parenthesis like `%(name)s`, as this are variable names

PR #2663, opened 2023-01-17
PdfFileReader -> PdfReader

### LINKED ISSUE #2659: PDF meta data reader is not working
image

This is from a freshish (~2 weeks) install from master. 

The requirements note `PyPDF2>=1.20,<3.1.0` but the pdf meta data reader is apparently using a feature that was removed on v3.0. We should look to reduce the requirements, or better yet fix the meta data reader

blitzmann:
Fixes #2659

### DIFF
--- a/cps/uploader.py
+++ b/cps/uploader.py
@@ -37,12 +37,12 @@ except (ImportError, RuntimeError) as e:
     use_generic_pdf_cover = True
 
 try:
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Thanks. I increased the lowest working version of pypdf2 to 1.28, as the file reader class was renamed in this version, with older version it wasn't working

PR #2672, opened 2023-01-24
CSP invalid syntax

Petipopotam:
CSP requires some cosmetic corrections

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -85,11 +85,11 @@ def add_security_headers(resp):
     csp += " 'unsafe-inline' 'unsafe-eval'; font-src 'self' data:; img-src 'self'"
     if request.path.startswith("/author/") and config.config_use_goodreads:
         csp += " images.gr-assets.com i.gr-assets.com s.gr-assets.com"
-    csp += " data:;"
+    csp += " data:"
     if request.endpoint == "edit-book.show_edit_book" or config.config_use_google_drive:
         csp += " *;"
     elif request.endpoint == "web.read_book":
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Could you please tell me, the problem which this correction solves.

Petipopotam:
Of course.

The current CSP is : 

default-src 'self'  'unsafe-inline' 'unsafe-eval'; font-src 'self' data:; img-src 'self' **data:** style-src-elem 'self' blob: 'unsafe-inline'**;object-src:** 'none';

It's invalid : 
1/ a semicolon is missing to close the img-src statement (just after data:) : it means that the img-src statement is invalid, the blob: value is not applied to img-src and then web reading ebook is not displaying image.

2/ there is a colon just after object-src but CSP directives don't end with a colon

You can check that with Chrome debugger and you will see error notified or you can check the current CSP with https://csp-evaluator.withgoogle.com/

The patch I suggest modify this CSP to properly close img-src statement after adding the blob: value for that directive (required to display image for web reading ebook) and remove the colon for the object-src statement.

default-src 'self'  'unsafe-inline' 'unsafe-eval'; font-src 'self' data:; img-src 'self' data: **blob:;** style-src-elem 'self' blob: 'unsafe-inline'; **object-src** 'none';

Petipopotam:
I'm not sure that blob: value is required for the style-src-elem directive here but I kept that unchanged

Petipopotam:
Capture d’écran de 2023-01-30 10-27-05

Issue #2694, opened 2023-02-14
Using book title rather then file name when sending book to device


huyumars:
## Proposal
Current, when you send book to device (via email), the attachment name is book file name:
>  Shi Jie Shang Zui Jian Dan De Hui Ji Shu - Darrell Mullis.mobi

Which has been converted English characters for better compatibility.
However it is not friendly for non-English user,  so we hope it could use book title stored in Calibre DB, like:

> 世界上最简单的会计书.mobi

## Solution
Using book title as attachment file name

PR #2695, opened 2023-02-14
use book title not file name as attachment name when sending email

huyumars:
#2694


bread22:
I love to have this option as well, any plan to get merged?

OzzieIsaacs:
Thanks for the contribution. In calibre-web there is a setting `config_unicode_filename` which should lead to the right filename if not set. I agree with you that "old" files stored in the English name expression are downloaded in the "old" style. 
I'd like to ask you to respect/apply this setting before converting the file name

PR #2701, opened 2023-02-21
Show stars in custom rating column

KelsierLuthadel:
When viewing custom columns in book details, ratings show as values. This fix shows the ratings as stars


OzzieIsaacs:
I'm not fully happy. The current ratings setting allows only "full stars" from calibre itself. The custom column also allows half stars, this isn't respected by your change. 
Furthermore it does not change the edit screen, this is from my point a bit confusion for users, you edit numbers but in another screen you then see stars. 
I'd like to ask you to make is complete. The current "star editing js script" is not able to edit half stars, so my suggestion would be to use a single select to present the different selections

Issue #2371, opened 2022-04-14
Kobo sync a book from the kobo store doesn't download the cover


JVT038:
**Describe the bug/problem**
So I bought a book from the kobo store and I wanted to sync it.
I checked the 'Proxy unknown requests to Kobo Store' checkbox in the basic configurations, and my 'server external port' is set on 80.
The book is downloaded like it's supposed to, but it doesn't load / download the cover.

**Environment (please complete the following information):**
 - OS: Synology Diskstation Manager 7
 - Python version: 3.8.10
 - Calibre-Web version: 0.6.19 Beta - df53a5d8c97fe6790f75ed1d9e073b5f614dfc01 - 2022-04-13T18:37:23+02:00
 - Docker container: LinuxServer:
 - Special Hardware: Synology NAS DS918+
 - Browser: Firefox

**Additional context**
I access calibre-web through a reverse proxy

jccpalmer:
Hey, did you ever get this resolved? I'm experiencing the same problem. I don't get covers from the store or Overdrive.

JVT038:
Nope, no solution.

flipside101:
Just wanted to add in im seeing this as well

karlshea:
Looks like #2731 would fix this

johnnyfleet:
I had the same issue and can now confirm that the above fix seems to have worked. Running [linuxserver/calibre-web:latest](https://hub.docker.com/r/linuxserver/calibre-web) docker instance and all works well now. 

Might be worth checking now @JVT038 .

karlshea:
I don't think that docker image has the latest commits, when I look at contents the changes from 2731 aren't in there yet. I think there would need to be a new tagged release.

ciceroii:
I'm running the most recent version, and still having the problem of the covers not showing. Is there anyway to debug why they might not showing?

karlshea:
I have the new version and mine are also not working.

rtrox:
It looks like Kobo changed their image endpoint, and this is currently broken again. From a fresh kobo sage's config:
```
image_host=https://cdn.kobo.com/book-images/
image_url_quality_template=https://cdn.kobo.com/book-images/{ImageId}/{Width}/{Height}/{Quality}/{IsGreyscale}/image.jpg
image_url_template=https://cdn.kobo.com/book-images/{ImageId}/{Width}/{Height}/false/image.jpg
```

rtrox:
Opened a PR to resolve.

[...]

PR #2731, opened 2023-03-28
Fix proxy cover images to Kobo store

tader:
Proxying cover requests to the Kobo store failed because of way generic covers could be generated for missing covers.

An option `use_generic_cover_on_failure` was previously introduced to disable generic cover generation and return a 404 status code instead, if set to `False`. Because of that 404, the redirect to the Kobo store was never reached.

I've removed all the `use_generic_cover_on_failure` handling, performing the redirect or 404 in the `HandleCoverImageRequest` method instead.

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -747,9 +747,10 @@ def get_book_cover(book_id, resolution=None):
     return get_book_cover_internal(book, use_generic_cover_on_failure=True, resolution=resolution)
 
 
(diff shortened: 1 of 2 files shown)


flipside101:
fingers crossed this get recognised and merged

johnnyfleet:
Adding my +1 here too. Looking forward to this getting attention and hopefully merged. Seeing same behaviour as #2371 so hopefully this fixed that.

flipside101:
@OzzieIsaacs thankyou

PR #2736, opened 2023-04-01
Implement fetching metadata from mp3 files

hennadii-demchenko:
Here's a feature I'd like to have when importing audiobooks in MP3 format. 

A couple of notes/questions:
- your repo declares support of python3.5+, but `eye3d` which I used, states it supports 3.7+, would this be ok?
- the code here is pretty basic, should I also consider writing tests to it, or is it good as it is?

### DIFF
--- a/cps/mp3.py
+++ b/cps/mp3.py
@@ -1,0 +1,32 @@
+from eyed3.core import Date
+from eyed3.id3 import Tag
+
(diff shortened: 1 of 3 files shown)


OzzieIsaacs:
I decided to use "mutagen" to fetch metadata from more than only mp3 files.

PR #2744, opened 2023-04-10
Added book randomization functionality next to sorting buttons

maciekpoplawski:
Added random sort button as per my wife request to be able to randomize books in certain categories. I think it would be beneficial for people who can't make their mind on what to read :)


OzzieIsaacs:
Firsthand this sounds like a plan, the problem is with your code that each page will create a new database query with random sort order, so it's possible that books repeat itself from each page request to the other and no real random order is archived.
There is the "discover books section" which does exactly the same as your code except for the choice a certain category/series/publisher..., but it is limited to one page. Furthermore your code has the problem that it is missing the random function for at least authors and search results, maybe also shelves (I'm not sure if it makes sens in this case).
So I decided not to merge it

PR #2769, opened 2023-04-23
Nxc chinese

gushedaoren:
Solve Chinese garbled code
 # Solve Chinese garbled code
from urllib.parse import unquote
response.headers["Content-Disposition"] = "attachment; filename*=UTF-8''{}".format(quote(data.name))


OzzieIsaacs:
Could you please give me an example what exactly the problem is?

gushedaoren:
> 

When I Download Chinese book ,the book name is cannot readable and the download file cannot open by bookreader.
WechatIMG197
WechatIMG198

OzzieIsaacs:
Almost the same header as yours is created in https://github.com/janeczku/calibre-web/blob/de59181be714e2d87a62d4b7d4fc83162145c120/cps/helper.py#L1046
I tested several browser Firefox, Chromium, Safari on Linux, Windows, iOs all of them got the right name, so my conclusion is, your bookreader app has a problem not the code.
I removed the encode utf in the latest commit, but I don't see a reason to change more of the code

PR #2772, opened 2023-04-26
Quick search improvements

quarz12:
Currently the quick search bar only allows searching for either title or author, not a combination or different word order.
With this pr, author and title can be searched together and word order does not matter anymore.

Ideally I want to implement a way to ignore typos as well.

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -889,7 +889,7 @@ class CalibreDB:
         return self.session.query(Books) \
             .filter(and_(Books.authors.any(and_(*q)), func.lower(Books.title).ilike("%" + title + "%"))).first()
 
-    def search_query(self, term, config, *join):
+    def search_query(self, term, config, *join)->sqlalchemy.orm.Query:
         term.strip().lower()
         self.session.connection().connection.connection.create_function("lower", 1, lcase)
         self.session.connection().connection.connection.create_function("partial_ratio", 2, partial_ratio)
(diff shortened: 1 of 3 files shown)


quarz12:
I think this is now ready for review. I added fuzzy matching for all search categories but custom-columns. I was not able to understand what they do, so search behavior for those remains unchanged. Also, this change does not affect the advanced search feature. Please let me know if there's anything to clear up.

quarz12:
I found some issues with longer queries. For now this is still WIP

quarz12:
Books now also get sorted based on how close they are to the search parameters. Words < 3 letters get ignored in both the search query and book attributes when filtering. If the query does not contain words longer than 3 letters, no results are returned.
This PR is now again ready for review. 
However there are two more things to mention here. 
1. when a query returns 0 results, theres now a message showing "Words smaller than 3 letters are not considered". I did not know where to put its css, so for now it's left-bound.
2. I am not sure how this feature should interact with the so far existing sorting options. For now it seems to just override the default, time based, newest first, sorting. There probably should be a new button for this sorting option to still allow time based, newest first sorting.

quarz12:
@OzzieIsaacs Have you had time to look at this yet? What do you think?

kanjieater:
Would this PR also address issues where Calibre search returns completely different results than Calibre Web?  https://github.com/janeczku/calibre-web/issues/2805

Seems like calibre web might be filtering out works that doen't have the same columns.

quarz12:
Given that this search version is a lot more forgiving, I'd say there's a good chance it does at least improve that issue

quarz12:
@OzzieIsaacs just want to remind you that this PR exists. I think this would be a big improvement to the current search

PR #2778, opened 2023-05-08
fix for home button linking to server root instead of application root

quarz12:
I changed the link target of the home button so that it now always links to the calibre-web root. Before, if hosted on a server, it would link to the server root. Like if the app was on www.mypage.com/calibre-web it would link to www.mypage.com

Also, I fixed a typo in the github report link of http_error.html.


Lava-Goat:
Please accept this request. I was just going to write one of my own after upgrading.

quarz12:
@OzzieIsaacs It's been a while, just want to make sure this won't be forgottten

Issue #2765, opened 2023-04-21
Read percentage for epub reader


gavdgavd:
I would like to have a ‘percentage read’ indication on the epub reader. For example, Calibre has this feature in their reader which displays a small number plus “%” on the right-bottom corner of the window.
Note: This is not an issue, but a Feature request.

OzzieIsaacs:
Is meanwhile implemented

PR #2784, opened 2023-05-15
EPUB progress indicator

quarz12:
this draft aims to implement #2765
My idea for achieving this is to compare the bytesize of all html files to the ones already read +the read percentage of the current book. while this does not give a 100% accurate representation as it counts the entire xml-tree instead of just the text fields , I think for this purpose it is more than good enough.


quarz12:
I found out that the percentage calculation could be done with the already included epub.min.js lib. This has the one downside that it has to build an index on load and therefore takes a few seconds to show the correct position when opening a book. However it means the other two libs I added before could be removed and the implementation is a lot simpler

Issue #2763, opened 2023-04-21
Google OAuth2 token_expired


oceanusxiv:
**Describe the bug/problem**
If the user's Oauth2 access token has expired, instead of prompting for the Oauth2 login again, calibre-web just outputs a 
```
[2023-04-20 19:56:06,246] ERROR {cps.oauth_bb:360} (token_expired) 
```
error and refuses to login.

**To Reproduce**
1. Setup Google Oauth2
2. Login with Google Oauth sucessfully
3. Wait until access token expires in browser (should take a day or two?)
4. Attempt to login again

**Logfile**
```
[2023-04-20 19:56:06,246] ERROR {cps.oauth_bb:360} (token_expired) 
```

**Expected behavior**
If the user access token has expired, login with Oauth should redirect to google again for re-authentication.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Python version: 3.8
 - Calibre-Web version: 0.6.20
 - Docker container: None
 - Special Hardware: Raspberry Pi 4
 - Browser: Safari

**Additional context**
Add any other context about the problem here. [e.g. access via reverse proxy, database background sync, special database location]

rickfmn:
I added a new branch to my fork of Calibre-Web to help fix this issue.  When you get the token expired message you would click on the "clear token" button on the login form.  This will clear the session and token cookies.  And it should work again.  

https://github.com/rickfmn/calibre-web/tree/FixExpiredToken

Let me know if you have any issues on it.

PR #2843, opened 2023-07-24
Fix expired token

rickfmn:


OzzieIsaacs:
This PR is referring to #2763, right?

Please use url_for function to generate the path, otherwise the logic will not work behind proxies with path suffix
If I see it right the clear token button is also visible if a user doesn't use oauth, this should not be the case

Miguel736j (approved):

PR #2871, opened 2023-08-24
Add support for systemd socket-activation

PhracturedBlue:
This is a small patch to support systemd socket-activation on Linux.

This allows on-demand startup of calibre-web on the first socket connection, which allows me to save resouces on a service that is not in continuous use.

Sockets can be either AF_INET or AF_UNIX, and will the auto-detected

### DIFF
--- a/cps/server.py
+++ b/cps/server.py
@@ -27,6 +27,7 @@ try:
     from gevent.pywsgi import WSGIServer
     from .gevent_wsgi import MyWSGIHandler
     from gevent.pool import Pool
+    from gevent.socket import socket as GeventSocket
     from gevent import __version__ as _version
     from greenlet import GreenletExit
     import ssl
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
The fix looks simple. But I don't have expertise in this "field of linux".

Could you please give me a working example? 
What do I have to do to use this code?
How can I test this?

PhracturedBlue:
It can be used like this:
Create a calibre.socket file in /etc/systemd/system/calibre.socket:
```
[Unit]
Description=calibre

[Socket]
ListenStream=5555

[Install]
WantedBy=sockets.target
```
And a service file in /etc/systemd/system/calibre.service that looks like this:
```
[Unit]
Description=CalibreWeb
Wants=network-online.target
After=network-online.target

[Service]
Restart=on-failure
TimeoutStopSec=70
ExecStart=cps
Type=notify
NotifyAccess=all

[Install]
WantedBy=default.target
```
then after enabling in systemd (`systemctl daemon reload && systemctl enable calibre.socket && systemctl start calibre.socket`)
Now Calibre-Web is not running, but will start as soon as a browser connects to localhost:5555
Note that I don't need to tell calibre-web what port to listen on.  Systemd is opening the port and forwarding the fle-descriptor to calibre-web

My personal use is a bit more complicated (I use a unix-domain-socket in the socket file, and then have nginx forward data to that socket, and i run calibre-web in a Podman container), but the use-case is the same (delay application startup until 1st use)

For posterity, a full example of my use-case (with dockerfile and systemd files) can be found here: https://github.com/PhracturedBlue/podman-socket-activated-services/tree/main/calibre-web

OzzieIsaacs:
Thanks for the unit files. I merged the code and also added the function for tornado as server

PR #2883, opened 2023-09-08
Add Favicon to OPDS Feed

databoy2k:
"< icon >" is a standard of the OPDS spec, and this commit adds the favicon to the feed. Certain OPDS readers, e.g. FBReader on Windows, respect this tag and will show the same icon as the main page.


OzzieIsaacs:
You did use a absolute path to the favicon, I changed this

Issue #2881, opened 2023-09-07
Suggestion: Permit send by email to E-reader even if download is prohibited


Normand-Nadon:
**Feature:**  
It would be great if the send to e-reader feature was separated from the "allow downloads"

**Use case:**   
We use calibre web for a small community of about 20 people. It is used as a community library.
I don't want to give permission to the users to download the books directly on the interface, but we do lend Kindle E-readers.
I would like the users to be able to send the book they want to read on those E-readers.
(For Kobos, we have a suitable workaroud already)

Proposed solution:
- Make allow downloads and email share 2 different permissions
- Make email sharing address manageable only by admin, or by the user too

Alternatives considered
We already have a couple of Kobos, using a different account. The synking process on Kobo makes it simpler.
Truth is that Kindle tablet are a lot simpler to obtain and cheaper.

**Additional context**
Your project is amazing! keep-on the good work!

apollo1220:
Image containing this changed is published to apollo1220/calibre-web on Docker Hub

PR #2892, opened 2023-09-19
Add send to ereader permission

apollo1220:
Implement suggested feature [2881](https://github.com/janeczku/calibre-web/issues/2881)
Add send to ereader permission and separate the download and send to ereader button visibility check to check the individual permissions instead of both keying off the download permission.


apollo1220:
Image containing this changed is published to apollo1220/calibre-web on Docker Hub

OzzieIsaacs:
I will not split the permission further, this adds a lot of complexity

PR #2211, opened 2021-12-08
feat: support generic oauth2

decentral1se:
Hi following https://github.com/janeczku/calibre-web/issues/2199 I've been looking into the feasability of meeting a compromise between 1. not wanting to add additional login methods and 2. supporting a broader set of oauth2 providers.

My solution for this has been to implement a "generic" oauth2 provider which can support anything that supports oauth2. This would mean that you could potentially deprecate Github/Google methods and use this Generic method as the one oauth2 implementation. All sorts of oauth2 providers can then be supported without further coding work.

I don't think my implementation is very good but it is a start and it does indeed work. I'm testing it now. I don't know how to support migrations so this breaks when running with existing installations.

Curious if this can be merged in some fashion or if there is space for more discussion?

PR #2899, opened 2023-09-26
allow creating users from reverse proxy headers

igor47:
Currently, when authentication happens in the reverse proxy, the user must already be in `calibre-web`'s DB or else nothing happens. [Source](https://github.com/janeczku/calibre-web/wiki/Setup-Reverse-Proxy#login-via-header-from-upstream-authentication-source):

> If you pass a username that isn't present in the database, nothing will happen - the user must exist beforehand in order to login.

With this PR, we add an option to also create the user. User creation is modeled on how we do it in LDAP sync. If the user creation fails for whatever reason, still nothing happens, and the failure happens transparently to the user, exactly how it happens today if the user doesn't already exist in the DB.

This is a very-low code PR, but it makes authenticating with reverse proxies *much* easier. The alternatives to this PR are:
* use reverse-proxy auth, but create all users manually -- this is tedious, I doubt anyone does it
* use LDAP sync to create users, and then authenticate via the proxy -- this is the most sensible alternative, but is a lot of work since LDAP is a hot old mess.
* wait for, or resume, work on the oauth generic provider from #2211

i think for people using centralized auth (authentik/authelia) with a reverse proxy (traefik/caddy/nginx) this provides the biggest bang-for-buck without much code to maintain.

We don't do any group setting here, so users will likely be created non-admin. My recommended flow to set up auth:
* set up calibre-web using the docker images
* log in with the default admin user and set up reverse proxy auth settings
* create a user matching the name you'll get from your proxy, and make that user an admin
* enable auth in your proxy
* log in with your proxy auth provider. you should be an admin
* delete the default admin user

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -1846,6 +1846,8 @@ def _configuration_update_helper():
         # Reverse proxy login configuration
         _config_checkbox(to_save, "config_allow_reverse_proxy_header_login")
         _config_string(to_save, "config_reverse_proxy_login_header_name")
+        _config_checkbox(to_save, "config_reverse_proxy_create_users")
+        _config_string(to_save, "config_reverse_proxy_email_header_name")
 
         # OAuth configuration
         if config.config_login_type == constants.LOGIN_OAUTH:
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -154,6 +154,8 @@ class _Settings(_Base):
 
     config_reverse_proxy_login_header_name = Column(String)
     config_allow_reverse_proxy_header_login = Column(Boolean, default=False)
+    config_reverse_proxy_create_users = Column(Boolean, default=False)
+    config_reverse_proxy_email_header_name = Column(String)
 
     schedule_start_time = Column(Integer, default=4)
     schedule_duration = Column(Integer, default=10)
--- a/cps/templates/admin.html
+++ b/cps/templates/admin.html
@@ -154,6 +154,14 @@
             <div class="col-xs-6 col-sm-7">{{_('Reverse Proxy Header Name')}}</div>
             <div class="col-xs-6 col-sm-5">{{ config.config_reverse_proxy_login_header_name }}</div>
         </div>
+        <div class="row">
+            <div class="col-xs-6 col-sm-7">{{_('Create Reverse Proxy Users')}}</div>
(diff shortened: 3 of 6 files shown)


MisterMustache:
Why hasn't this been merged @OzzieIsaacs?

igor47:
FWIW, i've been running my own fork of calibre-web with this feature since i opened the PR, with no problems.

MisterMustache:
> BTW @MisterMustache -- your comment came off as pretty aggressive to me, not sure if that was the intent, but i prefer to be more gentle to overworked OSS maintainers.

God no! I'm sorry if It came out that way -- that was was never my intention. I made the comment with genuine curiosity because I haven't seen the README with the note you posted in the comment. I fully support the main maintainer and wish him all the best.

> FWIW, i've been running my own fork of calibre-web with this feature since i opened the PR, with no problems. this PR is open for the benefit of the community. here's how i run it:
> - [docker-compose](https://github.com/Moomers/mittens.services/blob/main/docker-compose.yml#L201-L225)
> - [dockerfile](https://github.com/Moomers/mittens.services/blob/main/custom/calibre/Dockerfile)

Thank you for contributing, I really appreciate it! 😄 
I've been thinking of making my own fork as well. That being said the links you provided are not public :( and return a 404 to me.

nOw-Ay:
Hello @igor47, just had an issue with your PR. I am using Keycloak and Traefik (getting reverse proxy auth thanks to [this bridge](https://github.com/thomseddon/traefik-forward-auth)), I set up `X-Forwarded-User` as the user header and left the email field blank. Despite my vigilance, I got this error :
```
  File "/app/calibre-web/cps/usermanagement.py", line 139, in create_user_from_reverse_proxy_header
    user.email = email
UnboundLocalError: local variable 'email' referenced before assignmen
```
Do you have an idea of what might not be working ? Thanks a lot.

igor47:
> I set up `X-Forwarded-User` as the user header and left the email field blank.

yup i wasn't handling this. i updated the PR.

aroberts:
Any hope for this to get merged? I'm returning to calibre-web after a long break (I originally contributed the header-based login code), and this would be quite nice to have. @janeczku are there any issues you see with this PR that keep it from being merged? If it's abandoned, I'd be willing to fix the conflicts.

igor47:
i rebased the PR

Issue #732, opened 2018-12-25
[Enhancement] Scrolled reader


Oreolek:
Can you add an option to use scrolled epub reader, not spreads? Spreads are kinda buggy on small screens and scrolling is easier to use on desktop overall.

OzzieIsaacs:
The used reader javascript is capable of presentinhg a scrolled reader. I‘ll check how difficult it will be to integrate

sss-ng:
I'd love this feature. 
I'm curious what happened with PR #2902 
Was it closed because it didn't work as expected or does the maintainer not want this feature?

PR #2902, opened 2023-09-30
Add epub reader flow setting to allow scrolled rendering

wbradley93:
Implements suggested enhancement #732

Currently, the web ePub reader only allows paginated rendering of documents. This PR adds a dropdown to the settings modal that allows the user to switch to a scrolled rendering of the current chapter.


wbradley93:
Hi all, anything I can do to push this PR along? Thanks!

PR #2909, opened 2023-10-13
Fix url validation and author name

deldesir:
URL validation is better handled taking into consideration short ones without the protocol.
Author name are added properly in metadata.


deldesir:
Apologies. PR wasn't intended for upstream.

PR #2926, opened 2023-11-09
Save progress to database on every page change in epub reader

dydyamotya:
Hello!

My girlfriend wants to read the book both on her smartphone and laptop, and she wants it to be synchronized between both devices. So, I made this addition to online epub reader.

I also saw a couple of requests for feature before on the net, so this could be popular improvement.

Please check if it is ok for the project. 

Thank you for your attention.

### DIFF
--- a/cps/static/js/reading/epub.js
+++ b/cps/static/js/reading/epub.js
@@ -10,7 +10,8 @@ var reader;
 
     reader = ePubReader(calibre.bookUrl, {
         restore: true,
-        bookmarks: calibre.bookmark ? [calibre.bookmark] : []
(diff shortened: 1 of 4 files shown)


sk4rd (approved):

1447262390:
This feature is very useful, thank you

linonetwo:
will this work for txt? #2946

Zwixx:
> will this work for txt? #2946

No, it only works with epub

OzzieIsaacs:
I don't think this is good idea, as it creates a lot of write actions on the database (I will also revert this for the comic reader)

PR #2997, opened 2024-02-05
Feature: Sync custom column read state with kobo

### LINKED ISSUE #2740: Sync Read State between Kobo, Calibre-Web and Calibre Custom Column
If "Link Read/Unread Status to Calibre Column" is set, the Calibre-Web will not show the book marked as "Finished" on the Kobo reader as read and if the setting is not set it will reflect the Kobo state, but not the Calibre Custom Column.

Sync all three read states. Meaning that when a book is finished on the Kobo it will be marked as Read in Calibre-Web and marked in the Calibre Custom Column

Setting or unsetting the Custom Column in Calibre-Web and makring the unsynced state manually

**Additional context**
---

NotSimone:
Currently, the read state when books are synced with Kobo is not correctly synced when the "Link Read/Unread Status to Calibre Column" option is used.

This change makes it so that changing read state on either Calibre or the Kobo e-reader should sync the read state correctly. Note that the Kobo reading state has a third state - reading - that cannot be represented in the Calibre column and so gets ignored.

This also adds an additional configuration option "User for Calibre Read Column" so that only the specified user's syncing will update or be updated by the column in Calibre. By default, it is configured to be the Admin user (id 1).

image

resolves #2740

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -246,8 +246,11 @@ def db_configuration():
 @user_login_required
 @admin_required
 def configuration():
+    all_user = ub.session.query(ub.User)
+    all_user = all_user.filter(ub.User.role.op('&')(constants.ROLE_ANONYMOUS) != constants.ROLE_ANONYMOUS)
     return render_title_template("config_edit.html",
                                  config=config,
+                                 users=all_user.all(),
                                  provider=oauthblueprints,
                                  feature_support=feature_support,
                                  title=_("Basic Configuration"), page="config")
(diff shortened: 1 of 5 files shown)


NotSimone:
I should mention a caveat - it seems that edition the custom column in calibre does not change the last_modified timestamp (or anything else vaguely resembling a timestamp afaik), which means you cannot reliably sync a change in read state from Calibre to Kobo, but the other way is fairly reliable. Fortunately, that's the direction most of us care about anyway ;)

quizzmaster:
I merged this PR for myself and it works like a charm :)

NotSimone:
I've just rebased this onto the tip of develop

sgtio:
This would be a great thing to have! I was going crazy wondering why Calibre web was not setting the read column in my Calibre library after I finished a book in my Kobo. But if I got it right, the column is only written to if you read a book through Calibre web? Or is it only read by Calibre web, but never written to? 

Anyway, any hope that this PR can be merged?

NotSimone:
The custom column is already being updated properly when the read state is updated through the web UI. This change only affects the column state when it is synced against a Kobo e-reader.

quizzmaster:
@janeczku Can this get merged? This PR is open for almost 2 Years, is quality of life for all that use a custom columns for the read state and introduces very few changes.

Issue #2723, opened 2023-03-18
Calbire-Web - Kindle friendly browser interface


IdefixRC:
**Is your feature request related to a problem? Please describe.**
I currently use calibre and the included calibre content server on a windows machine to allow the various kindle devices in my house to directly download ebooks using the kindle web browser and access the non java page of the calibre server (/mobile).
The formatting of the mobile page works great for the kindle.
I'm now experimenting with calibre-web running in a docker container. it works great but accessing content via the kindle browser is way more troublesome than calibre server as the GUI of calibre-web does not format well on the kindle (way to small fonts, full calibre-web function etc) making downloads a bit cumbersome.

**Describe the solution you'd like**
I might be missing something but it would be great to have a calibre server (mobile version) type of interface option for calibre web to allow easier content download utilizing the kindle browser.

**Describe alternatives you've considered**
There is a calibre docker container which runs the standard calibre SW but it only works with a number of workarounds (e.g. the entire server needs to be restarted each time new books are added from remote calibre instances.

Thanks !

IdefixRC:
Update: went with COPS (https://github.com/seblucas/cops) for now as I could not find a simple web interface for the kindle through calibre-web. Would be great to see something similar in calibre-web in the future.

pnatel:
+1 on a kindle friendly version. 
Maybe just add the option to request the mobile version or the web version.
or add a /mobile to the site like Calibre.

fferraro87:
+1

carderne:
Yes I can experiment with that, a small-ish image should be no problem.
I can also add a tiny bit of CSS back just so it looks a bit nicer, but still optimised for the screen size etc.

Will update here when I get to it.

carderne:
@pnatel 
Have updated my fork a bit:
1. Improved CSS
2. Added images
3. Added pagination

Give it a go by running the new Docker image [carderne/calibre-web:0.6.21-2](https://hub.docker.com/layers/carderne/calibre-web/0.6.21-2/images/sha256-57ca32b4ea4cf4802ceaf9cf7bb4a7dc21e53c2285074134dc489707423264a1?context=explore).

Or pull my repo [carderne/calibre-web](https://github.com/carderne/calibre-web) and do it manually.

pnatel:
Thanks, @carderne, 
[...]

PR #3008, opened 2024-02-28
Add Kindle-friendly browser interface

carderne:
## Relevant issues
- https://github.com/janeczku/calibre-web/issues/2723

## Background
It would be great to be able to use calibre-web from a Kindle and directly download books. My current flow is to use calibre-web to manage/explore books, and then a separate process to get the book onto the Kindle.

## Description
This PR adds two new routes:
- `/basic?query=<term>`: equivalent of `/search`
- `/basic_book/<id>`: equivalent of `/book/<id>`

They have no CSS or JavaScript and very basic structure. This is aimed at creating a page that the Kindle web browser can easily handle.

The only features that are maintained:
- login (this page is unchanged)
- logout from every page
- simple search
- book list with `<authors> - <title>`
- book detail view with download links

## Comments
I went as basic as possible to create a small diff, demonstrate the purpose, and create a tiny maintenance surface area if this gets merged.

Some functionality could be added back, but it's a trade-off with the limited browser. Search, select and download is all I need, but others may have different requirements.

Obviously happy to implement this in a different way if you think preferable!

### Search page, after searching
home-screenshot

### Book detail page
detail-screenshot


carderne:
(Could also do this with by detecting the Kindle User-Agent, but that would involve changing the main routes, so I've left it with this light-touch method for now...)

DavidGriswoldTeacher:
Thank you for this @carderne . I am currently using the linuxserver docker image, can you explain to me how to switch to your version that adds the /basic path? I am relatively new to docker...

carderne:
Have updated my fork a bit:
1. Improved CSS
2. Added images
3. Added pagination and a home button

jamescochran:
This is great! Exactly what I was looking for. The search box is a bit wonky as well as logout appears under home but I assume it should be on the far right. But all in all it does what I was looking for and that was to grab a book from my Calibre library through the Kindle browser.

carderne:
@jamescochran 
I had only tested on the latest Kindle Paperwhite, just used an older one and realised it doesn't support flexbox...

Have updated my fork to only use old-school CSS, please try again!
(The Docker image is now `carderne/calibre-web:0.6.21-3` if that's what you're using.)

DavidGriswoldTeacher:
Very nice! I like it a lot. 

As a Kobo user who prefers this simple system over the more complex kobo syncing, I might at some point fork this myself and try to figure out how to add a link that will download epub in kepub format. I think the logic on this would be...

1. If the kepub is there, just include its link as normal, but with the logic necessary to convert the download link from .kepub to .kepub.epub (a requirement for the kobo browser to download it). This is a bit broken on the main code right now.

2. If the .kepub is NOT there, include a link that will convert the epub to kepub using kepubify, then download the  .kepub.epub file

this is not a feature request, just thinking "aloud"!

pnatel:
Same as @jamescochran with the only difference being that I don't have the login/logout option. I think it may be because I have the anonymous login enabled.
I also have some screenshots for reference in https://github.com/janeczku/calibre-web/issues/2723#issuecomment-2038092040

pnatel:
quick update, just installed carderne/calibre-web:0.6.21-3 and the search button is in the proper place.

The size of some of the elements looks off, but I guess that this is caused by the way the Kindle browser renders that type of object, or it is getting that from the main site.

carderne:
@pnatel can you share a photo of what you mean about "the search elements look off"? Maybe I can fix that.

carderne:
Oh damn ok I tested with an old-style Paperwhite and the new Paperwhite, but that's a third target. Not sure how I'll fix that without an Oasis to hand. Any chance you'd be able to run calibre-web locally and fiddle with the CSS?

The `<nav>` has a fixed `height: 50px` set, but Kindle's seem to treat `px` in mysterious ways... It might be as simple as setting it to `3rem` or something like that. Alternatively, set a height on the `<input` element as well maybe...

pnatel:
Created a pull request to your repo with small changes:
https://github.com/carderne/calibre-web/pull/1

adrianelsener:
Great, exactly what I was looking for!

carderne:
@pnatel your change is now available (along with other upstream updates from calibre-web) on Docker now as the following image:

[carderne/calibre-web:0.6.23](https://hub.docker.com/r/carderne/calibre-web/tags)

pnatel:
> @pnatel your change is now available (along with other upstream updates from calibre-web) on Docker now as the following image:
> 
> [carderne/calibre-web:0.6.23](https://hub.docker.com/r/carderne/calibre-web/tags)

Thank you very much!
I can confirm that it is working fine.

EnduringGuerila:
I really hope this gets merged in! I set up a server for friends and this would be a huge help, and I'd rather not have to set up a fork. Thank you for all the hard work!

carderne:
@OzzieIsaacs sorry to bother:
is there any chance of getting this (or something like it) merged, or should we put the effort in to automate building the fork?

OzzieIsaacs:
Thanks. This pull request has been integrated manually.

OzzieIsaacs:
Please use the url_for function for links in the future, otherwise links won't work with reverse proxy applications. I added a button in the normal theme to get to the basic theme (only visible if browser is detected as ebook reader)

PR #3063, opened 2024-06-01
Fix many italian strings

mapi68:


OzzieIsaacs:
Thanks.
There was one format string accidentally translated "(formato)%" I fixed it

Issue #3080, opened 2024-06-19
Forcing Bleach is causing problem displaying `<p>` tags


JeffersonQin:
The newest commit https://github.com/janeczku/calibre-web/commit/f7d5480811c72c7e8104e79a39d08b9407ae3338

Causing the following problem when editing metadata:

![image](https://github.com/janeczku/calibre-web/assets/30655701/15e5f8f0-29f2-454b-83cb-f261275e540a)

OzzieIsaacs:
Thanks for reporting and also for the fix. I have my own fix dveloped, I need one or 2 days to test (div, strong, pre and maybe more tags are also missing)

JeffersonQin:
I see. Thank you for letting me know.

OzzieIsaacs:
Should now fixed in the newest nightly version

PR #3081, opened 2024-06-19
fix: do not bleach `<p>` when cleaning

JeffersonQin:
Try to solve https://github.com/janeczku/calibre-web/issues/3080


OzzieIsaacs:
Could now be fixed in the newest nightly version

PR #3085, opened 2024-06-24
Possibility to rate from book page without editing

ofostier:
Not sure for the right CSS file to edit


ofostier:
Thanks for your feedback
I will try on docker setup

ofostier:
@AnonTester 
I finaly work on with a docker-calibre-web pointing to my dedicated release.
I have just found a bug concerning the old-rating as None value (fixed)

Except that, everything working fine on my side

ofostier:
Hmmm this is what I fixed 
Are you using this [release](https://github.com/ofostier/calibre-web/releases/tag/rating_0.2) ?

ofostier:
Strange cause the fix concening the error 500 is include in this patch
Can you retry

ofostier:
Thanks, I also found this error and fixed it (commit [c79a452](https://github.com/janeczku/calibre-web/pull/3085/commits/c79a452c54d09d74707e6eff457a3b29362bba63)).
We cannot unrating witout editing ebook (can try to add this feature soon)

Now I cannot get the Error 500 anymore but I cannot reproduce your new error concerning the Db constraint

ofostier:
Depending of the code owner

PR #3087, opened 2024-06-26
Allow using calibredb for upload processing

jgoguen:
Some things calibre is able to do via plugins that would be helpful to handle in calibre-web, but there's not much point in re-implementing those in calibre-web directly. Things like importing ACSM files and getting the resulting ebook directly, or importing KFX files that require extra processing. Currently, to keep the ability to use the Kobo sync service, the solution is to run both calibre-web and `calibre-server` and switch between them. Running `calibre-server` adds to smooth usability issues since using GUI calibre risks conflicting with `calibre-server` unless you remember to stop it every time you use GUI calibre. If your calibre library is stored on a shared folder, such as Google Drive or OneDrive or Dropbox, a conflict with `calibre-server` is guaranteed when using GUI calibre since the modified `metadata.db` is moved into place as a new file, with a new inode, causing `calibre-server` to be unable to continue using its existing database connection.

To accommodate using `calibredb` for uploads, avoiding conflicts between calibre and `calibre-server`, and allowing everything to happen in one interface instead of switching between calibre and calibre-web, this PR:

- Adds a new setting, linked to the main upload permission, to use calibredb for uploading files. Default value, or the behaviour if `calibredb` is not found, is to keep the existing behaviour of having calibre-web process the uploaded books and add them to the calibre database.
- Adds the location the linuxserver Docker image installs calibre to the list of paths searched for calibre binaries
- Adds the chunk of code to `upload()` in `editbooks.py` to shell out to `calibredb`
- Appends the original file extension to the temporary file path, since `calibredb` needs that for format detection

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -1763,6 +1763,7 @@ def _configuration_update_helper():
             return _configuration_result(_('Certfile Location is not Valid, Please Enter Correct Path'))
 
         _config_checkbox_int(to_save, "config_uploading")
+        _config_checkbox_int(to_save, "config_upload_with_calibredb")
         _config_checkbox_int(to_save, "config_unicode_filename")
         _config_checkbox_int(to_save, "config_embed_metadata")
         # Reboot on config_anonbrowse with enabled ldap, as decoraters are changed in this case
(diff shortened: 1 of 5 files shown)


OzzieIsaacs:
I thought some time on it, my conclusion is: I don't want to have an alternative upload processor, if someone wants to use calibre he can do it, but not using calibre-web. This adds up another level of complexity and will also confuse some users as there are already user which don't understand that calibre and calibre-web are 2 differnt programs.

PR #3113, opened 2024-08-02
Add bulk delete button in books table page

jmarmstrong1207:
This PR adds a button to delete selected books in the table page. Screenshots below show the workflow for end-users:

### **Select books**
image

### **Book deletion confirmation**
image

### **Result: Books deleted**
image

### DIFF
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -488,6 +488,28 @@ def simulate_merge_list_book():
             return json.dumps({'to': to_book, 'from': from_book})
     return ""
 
+@editbook.route("/ajax/simulatedeleteselectedbooks", methods=['POST'])
+@user_login_required
+@edit_required
+def simulate_delete_selected_books():
+    vals = request.get_json().get('selections')
+    books = []
+    if vals:
+        for book_id in vals:
+            books.append(calibre_db.get_book(book_id).title)
+        return json.dumps({'books': books})
+    return ""
+
+@editbook.route("/ajax/deleteselectedbooks", methods=['POST'])
+@user_login_required
+@edit_required
+def delete_selected_books():
+    vals = request.get_json().get('selections')
+    if vals:
+        for book_id in vals:
+            delete_book_from_table(book_id, "", True)
+        return json.dumps({'success': True})
+    return ""
 
 @editbook.route("/ajax/mergebooks", methods=['POST'])
 @user_login_required
--- a/cps/static/js/table.js
+++ b/cps/static/js/table.js
@@ -81,6 +81,14 @@ $(function() {
                 $("#merge_books").addClass("disabled");
                 $("#merge_books").attr("aria-disabled", true);
             }
+            if (selections.length >= 1) {
+                $("#delete_selected_books").removeClass("disabled");
+                $("#delete_selected_books").attr("aria-disabled", false);
+            } else {
+                $("#delete_selected_books").addClass("disabled");
+                $("#delete_selected_books").attr("aria-disabled", true);
+
+            }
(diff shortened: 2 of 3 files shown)


OzzieIsaacs:
I prefer to have the buttons for mass delete/unarchive in the top of the table, similar to the ones in the user edit table (.../admin/usertable):
image

jmarmstrong1207:
Looks like this now. It looks off; is this style good?
image

jmarmstrong1207:
When adding it into the table like above, table.js isn't enabling the buttons when there's at least 1 selections. It works when out of the table. Not sure why

jmarmstrong1207:
Okay, got it to work. Had to use `$(document).on('click', '#id', function....)` instead of `$('#id').click()` for the buttons because adding the buttons within the table header makes them dynamically-added elements.

Also added the bulk read/unread buttons. I have tested it to work, but of course testing yourself would be best 
image

jmarmstrong1207:
Alright, final change for this PR. I added bulk metadata editing:
image

jmarmstrong1207:
Let me know if I am missing anything. Fleshed out basically everything I can think of myself

cps/editbooks.py
```diff
@@ -350,6 +350,74 @@ def table_get_custom_enum(c_id):
 @edit_required
 def edit_list_book(param):
```
jmarmstrong1207, line 351: I moved the function to edit_book_param() in order to modularize it. This is so it can be used with the '/editselectedbooks/' calls. This essentially makes the change atomic from the client's perspective. Otherwise when bulk editing metadata, each metadata for each book needs an individual rest API call.

cps/kobo_sync_status.py
```diff
@@ -54,10 +54,10 @@ def remove_synced_book(book_id, all=False, session=None):
 def change_archived_books(book_id, state=None, message=None):
     archived_book = ub.session.query(ub.ArchivedBook).filter(and_(ub.ArchivedBook.user_id == int(current_user.id),
                                                                   ub.ArchivedBook.book_id == book_id)).first()
-    if not archived_book:
+    if not archived_book and state == True:
```
jmarmstrong1207, line 57: Makes it so only requests to archive (ie state = true) actually creates an archivedbook listing

cps/kobo_sync_status.py
```diff
@@ -54,10 +54,10 @@ def remove_synced_book(book_id, all=False, session=None):
...
         archived_book = ub.ArchivedBook(user_id=current_user.id, book_id=book_id)
 
-    archived_book.is_archived = state if state else not archived_book.is_archived
+    archived_book.is_archived = state if state != None else not archived_book.is_archived
```
jmarmstrong1207, line 60: Without this, bulk unarchiving on already-unarchived books will actually archive them instead. Without this, this function is just a toggler if state is false.

jmarmstrong1207:
Added ability to shift click to select multiple books at once. Works like in file explorer when shift clicking for multi-selection. Was a stupid easy change, just setting a property to True in the table creation

jmarmstrong1207:
@OzzieIsaacs This PR is ready to review now, whenever you're ready

jmarmstrong1207:
Do you know if this will be reviewed soon? @OzzieIsaacs

OzzieIsaacs:
I'll have a look at it. My request regarding buttons was a bit ambigious, the read/unread unarchive/archive ones should be radio buttons as on the user page. I'm feeling guilty for it, so I will change it by myself and I also will add the "select all" button. From the functionality it looks good

PR #3159, opened 2024-09-22
Korean translate update

abubaman:
Correction Typos.


abubaman:
I will re-pull request with messages.mo file.

Issue #2839, opened 2023-07-21
Bookmarks not working on tables


pocelka:
**Describe the bug/problem**
I created a bookmark (on any device) while reading the book. When I close the book and then open it again I can't display bookmarks while reading on tablet. I can access the bookmark on mobile phone / desktop, but on tablet I'm not able to click (touch) the bookmark icon on the left.

**To Reproduce**
1. Open book on iPad
2. Click on 'hamburger' menu on the left
3. Click on bookmark icon
4. Bookmarks not visible, bookmark list not showing

**Expected behavior**
It should be possible to see list of the bookmarks and click on them.

**Environment (please complete the following information):**
 - OS: iOA
 - Calibre-Web version: 0.6.20
 - Docker container: LinuxServer
 - Special Hardware: iPad 6th generation
 - Browser: safari, chrome, firefox

OzzieIsaacs:
PDF, epub, DJVU, txt, comic reader?

pocelka:
Epub

OzzieIsaacs:
In Firefox on Linux it's working, on iOS (checked 16.6 on iPad Pro), the bookmark button is showing no reaction

pocelka:
I have same iOS version (not sure about my tablet version, but it is an older model), but the bookmark button is not working for me. Just to be sure are you talking about the bookmark button (list) in the left menu right? Because I can create bookmarks just fine hitting the button in upper right corner, but then when I go for example few pages to the right and want to show "all bookmarks" I'm not able to click on that button (in left right menu). 

It feels like touch is not hitting the hit box. I tried also on FF and Safari.

OzzieIsaacs:
Yes, we are talking only on the ‚left‘ button

pocelka:
Than I don't know what could be wrong. If it is just on my side then I guess you need to classify this as "can't reproduce". I'm not sure if any one of my friends has apple tablet so I could try it also on with their devices.  I even tried to zoom in on the area (just to be sure that I touch the correct area) and still for me it's not clickable.

OzzieIsaacs:
?
I can confirm it's N O T working. I don't think there is a problem on your side.
?

Monoboy4ik:
[...]

PR #3178, opened 2024-10-08
fix bookmarks button for safari browser

lucolucus:
In the Safari browser, it was impossible for me to press the bookmarks button. After removing the `visibility: hidden`; property from `#panels a`, it started working again. 

Setting `visibility: hidden`; on the `#panels a` element makes the entire link (and all its content, including the` ::before` pseudo-element) invisible and not clickable.

I didn't have this problem on Chrome or firefox, and the modification didn't break anything for me.

 reference to  issue number #2839

### DIFF
--- a/cps/static/css/main.css
+++ b/cps/static/css/main.css
@@ -77,7 +77,6 @@ body {
 }
 
 #panels a {
-  visibility: hidden;
   width: 18px;
   height: 20px;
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Thanks for digging into it, it's also working for me afterwards, but still very buggy

PR #3196, opened 2024-10-22
Multiprocessing support

webysther:
This will reduce response time when multiple core cpu is available

### DIFF
--- a/cps/server.py
+++ b/cps/server.py
@@ -21,6 +21,7 @@ import os
 import errno
 import signal
 import socket
+import multiprocessing
 
 try:
     from gevent.pywsgi import WSGIServer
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
No it just throws an exception:
```
ERROR {cps.server:288} Error starting server: BaseAsyncIOLoop.start() takes 1 positional argument but 2 were given
Error starting server: BaseAsyncIOLoop.start() takes 1 positional argument but 2 were given
```
I checked the documentation of tornado back to 4.5, `start()` never accepted any parameter

webysther:
> No it just throws an exception:
> 
> ```
> ERROR {cps.server:288} Error starting server: BaseAsyncIOLoop.start() takes 1 positional argument but 2 were given
> Error starting server: BaseAsyncIOLoop.start() takes 1 positional argument but 2 were given
> ```
> 
> I checked the documentation of tornado back to 4.5, `start()` never accepted any parameter

This is running in my production as a patch, the tornado version you are using isn't +6?

https://github.com/janeczku/calibre-web/blob/master/requirements.txt#L11

PR #3199, opened 2024-10-23
Multiprocessing support

webysther:
Please check comments on #3196

### DIFF
--- a/cps/server.py
+++ b/cps/server.py
@@ -21,6 +21,7 @@ import os
 import errno
 import signal
 import socket
+import multiprocessing
 
 try:
     from gevent.pywsgi import WSGIServer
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
You are using gevent as server, so your code never gets executed.
The function doesn’t accept any parameters: 
https://www.tornadoweb.org/en/stable/_modules/tornado/ioloop.html#IOLoop.start
https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.stop

webysther:
> You are using gevent as server, so your code never gets executed. The function doesn’t accept any parameters: https://www.tornadoweb.org/en/stable/_modules/tornado/ioloop.html#IOLoop.start https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.stop

Oh! Now I get it, thanks!

PR #3212, opened 2024-11-01
Update Korean Translation

limeade23:
Hello.
I have improved the missing parts and awkward expressions in the Korean translation.


OzzieIsaacs:
Thanks

PR #3214, opened 2024-11-03
Reconnect to calibre database on SIGHUP

quantum5:
This PR hooks the "Reconnect Calibre Database" feature in the admin page to `SIGHUP`. This allows the database to be reloaded by running a command, which is extremely useful when using synchronization tools with the calibre `metadata.db`.

For example, Syncthing would delete `metadata.db` and recreate a new file in its place when the database is modified on a different machine, and  `inotifywait` can be used to send `SIGHUP` when this happens.

### DIFF
--- a/cps/main.py
+++ b/cps/main.py
@@ -42,6 +42,7 @@ def main():
     from .tasks_status import tasks
     from .error_handler import init_errorhandler
     from .remotelogin import remotelogin
+    from .signal import register_signals
     try:
         from .kobo import kobo, get_kobo_activated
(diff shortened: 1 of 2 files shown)


OzzieIsaacs:
I'm not a fan of coping/recreating/moving the metadata.db file while calibre-web is running. So I will not implement anything to make this easier. You could write a bash script with inotifywait and trigger the /reconnect route of calibre-web (https://github.com/janeczku/calibre-web/wiki/Command-Line-Interface#command-line-options, the -r option)

PR #3217, opened 2024-11-08
Add share details metadata

webysther:
This change allow share details on whatsapp and another app if 'Enable Anonymous Browsing' is enabled based on https://ogp.me/:

Before:

image

After:

image


OzzieIsaacs:
Not fully happy. I moved the code to the detail.html, so it's sure that there is an entry element and don't need to check it

OzzieIsaacs:
Please check if it's still working

webysther:
> Please check if it's still working

After upgrade on my side I will test, thanks!

PR #3223, opened 2024-11-13
Use correct mime type for fb2 files

Raf2k:
Both Moon Reader and Librera  (through opds) fails to show metadata and process fb2 books because of wrong file mime type. This small change fixes the issue.


OzzieIsaacs:
I doublechecked it, I converted a file with calibre to fb2 format. To upload it the mimetype has to match the file, it is identified as "text/xml":
grafik
So if the mimetype is changed the files are no longer uploadable

Raf2k:
You are right... I rushed and didn't properly test all the affected parts. Also I found that fb2 mime types can vary and not all readers will understand `text/fb2+xml` (for some of them it have to be `application/fb2`). Dolphin says it's `application/x-fictionbook+xml` So it's quite tricky... Unfortunately most part of my library is in fb2 format and I don't want to convert it. At least for now.  I'll think about it more... Thank you.

PR #3229, opened 2024-11-19
Using packaging library for more stable version checks. Tested with sqlalchemy

tejonaco:
Hi, because I'm on Arch Linux I realized that v0.6.24 crashes with sqlalchemy version (2.0.36.dev0) because of the int conversion of 'dev0'.
So I tried another approach using packaging library, It should be more strong for different version strings.
I only used it of sqlalchemy version comparison, but if it does work well it may be a good idea to use it for all packages.

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -24,9 +24,19 @@ import mimetypes
 import chardet  # dependency of requests
 import copy
 from importlib.metadata import metadata
+from packaging.version import Version
 
 from flask import Blueprint, jsonify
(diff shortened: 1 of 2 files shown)


OzzieIsaacs:
Thanks for pointing out the issue. I prefer to solve this more basic:
sqlalchemy_version2 = ([int(x) if x.isnumeric() else 0 for x in sql_version.split('.')[:3]] >= [2, 0, 0])
I will check if this line of code works and then close your PR

PR #3242, opened 2024-11-24
cps: Make --with-library optionally work with a Content Server

alimpfard:
This makes it possible to use the ebook conversion feature when an instance of Calibre is running, by allowing the user to specify the content server URL and credentials to perform updates with. Previously this action would make calibredb error out and fail the conversion job with an unknown error.

Ideally we'd also use this (if configured of course) to implement the book editing/upload/etc stuff, but those aren't broken for me currently :P


alimpfard:
Sure, up to you.
It's useful to me, but most people probably won't have my setup.

PR #3282, opened 2025-01-12
Fix German translation

tobiasbayer:


OzzieIsaacs:
Thanks for the translation, I agreed to most of the changes (some of my typos are a bit embarrassing) except:
The Guest Username is not translated, so I reverted it, but changed it to 'Guest' to make it more clear that it is a proper noun (Eigenname)

PR #3312, opened 2025-02-21
Refactor epub reader progress calculation

overhacked:
The code in `epub-progress.js` was causing the epub file to be loaded twice over the network, because it created a second instance of the epub.js library. I moved all the progress percentage display code into epub.js. Only the locationchange polyfill code was left in the separate file, renamed to `locationchange-polyfill.js`.

I rewrote the percentage progress code to use epub.js events, giving more succinct and readable code. Also, I added localStorage caching of the epub location calculations, required for displaying percentage progress, which can take up to 60-90 seconds to calculate on longer ebooks. This localStorage caching approach is [recommended by epub.js in the `locations.html` example in their repo](https://github.com/futurepress/epub.js/blob/f09089cf77c55427bfdac7e0a4fa130e373a19c8/examples/locations.html#L74-L84).


cps/static/js/reading/locationchange-polyfill.js
overhacked, line 1: I think that this polyfill might be unnecessary now. The previous percentage location update code used...
https://github.com/janeczku/calibre-web/blob/efc7ce31d232d8d094ce390ece5c50c56d466009/cps/static/js/reading/epub-progress.js#L41-L44
...to update progress, but this new code uses epub.js events:
```js
reader.rendition.on('relocated', (location)=>{
    let percentage = Math.round(location.end.percentage*100);
    progressDiv.textContent=percentage+"%";
});
```

So, maybe `locationchange-polyfill.js` can be removed altogether?

PR #3326, opened 2025-03-05
Adding ISBN, word count, and page count to Kobo sync

AsherMaximum:
Kobo devices can show the ISBN, word count, and page count on the "details" tab of the "view details" book page.

This change syncs the ISBN from the ISBN identifier to this field, and allows the user to link custom columns to the pages and word count fields.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -246,8 +246,11 @@ def db_configuration():
 @user_login_required
 @admin_required
 def configuration():
+    wordspages_columns = calibre_db.session.query(db.CustomColumns) \
+        .filter(and_(db.CustomColumns.datatype == 'int', db.CustomColumns.mark_for_delete == 0)).all()
     return render_title_template("config_edit.html",
                                  config=config,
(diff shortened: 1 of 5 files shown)


gelbphoenix:
Is it intended that the `custom_column_` entries for `config_kobo_words_cc` and `config_kobo_pages_cc` aren't created?

AsherMaximum:
No - the changes in config)sql.py should cause them to be created. 
When the app is launched, this function is run, which should create those columns if they do not exist.
```
def load_configuration(session, secret_key):
    _migrate_database(session, secret_key)
    if not session.query(_Settings).count():
        session.add(_Settings())
        session.commit()
```

It is intended though that they default to not being set to a particular custom column.

mirthquake:
FWIW, just wanted to chime in to say it would be great to have this functionality implemented in Calibre-Web. Am sorely missing the ability to view the word counts of books when picking the next one to read. Cheers!

PR #3332, opened 2025-03-13
Sort languages

thotmx:
Small fix for this issue: https://github.com/iiab/calibre-web/issues/303

I don't want to take the credit, this fix was created by (@deldesir)[https://github.com/deldesir], I'm only creating the PR to the upstream project.

@holta?


OzzieIsaacs:
There is more than one location where the locale is displayed (e.g. usertable), so I prefer to sort it in the backend.
This is done now in the latest nightly build

PR #3349, opened 2025-04-02
layout fixes

quarz12:
this pr removes this unused button:
image
on mobile resolution the author icon gets a fixed size to allow for a consistently centered placement
image
on small tablet resolution these buttons now behave the same (better than before but still not good). first picture is the old version, second one is new
image
image
and finally I removed some \<p> tags around a \<form> which is invalid html and can create issues due to browsers attempting to fix it automatically

### DIFF
--- a/cps/static/css/caliBlur_override.css
+++ b/cps/static/css/caliBlur_override.css
@@ -26,3 +26,36 @@ body.serieslist.grid-view div.container-fluid > div > div.col-sm-10::before {
 input.datepicker {color: transparent}
 input.datepicker:focus {color: transparent}
 input.datepicker:focus + input {color: #555}
+
+/* only mobile resolution */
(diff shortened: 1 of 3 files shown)


OzzieIsaacs:
I can't reproduce the 3rd behavior:
grafik

The 3 dots are not unused, they should allow to delete a single book format, this isn't working correct, but removing the 3 dots is not the solution:
grafik

quarz12:
you are looking at the phone resolution. It happens when the screenwidth is set to ~770 or more

diginfotek:
I could replicate the 3rd issue from 768px in width till around 1000px.

767px and below, it displays in mobile view.

Another fix option could be to move the Edit Metadata left panel on the top for resolution between 768px and 1000px?

PR #3350, opened 2025-04-02
hide desktop navbar when in mobile resolution

quarz12:
this hides the navbar in mobile resolution before it is moved to the collapsed sidebar to prevent this effect:
image


OzzieIsaacs:
I can't reproduce this with the newest nightly build and firefox or chromium:
grafik

It's looking good to me, do you have more details how to reproduce this?

quarz12:
you can see it if you throttle the browser to mobile speed. It only shows up for 1 second or so

PR #3351, opened 2025-04-02
change initial mobile book layout to reduce UI updates

quarz12:
this css change prevents the book table to shift the book icons once isotope is done ordering them which leads to less visible UI updates.
the last row now starts like this:
image
and not like this:
image


OzzieIsaacs:
Can't reproduce this (Firefox):
grafik

grafik

quarz12:
this is also something you need to throttle your browser for. I am using Chrome

PR #3352, opened 2025-04-03
Bugfix: fix some issues with kobo sync

xdgc:
Kobo sync's logic is incorrect, and looks like it pretty much always has been. This is seen on initial sync of a freshly reset device against c-w configured to sync all shelves. I have 1500+ books to sync, but c-w syncs only about 250 or so before saying it's done. That's two full batches of 100, followed by a partial batch.

The problem is in the way the query performs paging. On the first pass it discovers all the books that need to be synced. On the second pass it does pretty well.  Third pass, it does badly. All this is because of the way it currently detects which books have been modified since last sync. As it goes through the sync, this query result set gets progressively smaller, incorrectly.

I've tried to repair the logic, and what I get now works perfectly for my case. I'm not 100% sure it's exactly right but it's doing a lot better now for both initial sync and maintenance sync.

I had trouble understanding what was going on in kobo.py initially, so I also updated some of the logging to help out. This should be seen as a net improvement, though not strictly necessary.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -146,12 +146,19 @@ def HandleSyncRequest():
     if not current_user.role_download():
         log.info("Users need download permissions for syncing library to Kobo reader")
         return abort(403)
+
     sync_token = SyncToken.SyncToken.from_headers(request.headers)
     log.info("Kobo library sync request received")
     log.debug("SyncToken: {}".format(sync_token))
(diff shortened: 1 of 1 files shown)


xdgc:
@OzzieIsaacs  bump, this is a bug fix for basic functionality of the kobo OTA sync

PR #3353, opened 2025-04-03
Feature: keep c-w shelves in sync with calibre collections

xdgc:
This adds a new feature to sync Calibre-Web shelves to Calibre Collections. This can run ad hoc from a user's personal settings page OR as a scheduled daily task. Each user who has this feature enabled will be synced — users who do not individually enable it will not see changes.

When sync runs by either method, C-W detects which custom column from Calibre is configured as "collections" and then gathers all books that have collections set. For each named collection, if that collection does not exist as a shelf on the user's account, the shelf is created. Shelves are never removed. Then the set of bookids in the Calibre collection is diffed against the set of bookids in the C-W shelf. Missing books are added, and surplus books are removed. To maintain a shelf in C-W separately from Calibre, just name it something that doesn't exist in Calibre.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -56,6 +56,7 @@ from .usermanagement import user_login_required
 from .cw_babel import get_available_translations, get_available_locale, get_user_locale_language
 from . import debug_info
 from .string_helper import strip_whitespaces
+from .tasks.shelfsync import TaskSyncShelves
 
 log = logger.create()
 
@@ -480,6 +481,8 @@ def edit_list_user(param):
                     user.email = check_email(vals['value'])
                 elif param == 'kobo_only_shelves_sync':
                     user.kobo_only_shelves_sync = int(vals['value'] == 'true')
+                elif param == 'sync_from_collections':
+                    user.sync_from_collections = int(vals['value'] == 'true')
                 elif param == 'kindle_mail':
                     user.kindle_mail = valid_email(vals['value']) if vals['value'] else ""
                 elif param.endswith('role'):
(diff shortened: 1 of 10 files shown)


xdgc:
@OzzieIsaacs bump, please consider this compatibility feature

webysther:
Calibre collections are series in cw, how this will not cause a loop? If you sync cw shelf in calibre they will appear inside c-w as series, no?

xdgc:
> Calibre collections are series in cw, how this will not cause a loop? If you sync cw shelf in calibre they will appear inside c-w as series, no?

I'm afraid I don't follow. In Calibre, series and collections are different things. Calibre doesn't have native collections, but the Kobo Utilities and Kindle Collections plugins add it as a custom field (`#collections`). You can use this field to sync shelves on the Kobo, if you sync your Kobo with Calibre.

If you switch to using CW for Kobo sync, you lose that.

This feature update allows you to sync `#collections` from Calibre to CW shelves, so that CW can sync shelves to Kobo.

webysther:
> > Calibre collections are series in cw, how this will not cause a loop? If you sync cw shelf in calibre they will appear inside c-w as series, no?
> 
> I'm afraid I don't follow. In Calibre, series and collections are different things. Calibre doesn't have native collections, but the Kobo Utilities and Kindle Collections plugins add it as a custom field (`#collections`). You can use this field to sync shelves on the Kobo, if you sync your Kobo with Calibre.
> 
> If you switch to using CW for Kobo sync, you lose that.
> 
> This feature update allows you to sync `#collections` from Calibre to CW shelves, so that CW can sync shelves to Kobo.

Oh, now I get it, my bad.

PR #3355, opened 2025-04-04
Show shelf sizes in layout

xdgc:
Shows the number of items in each shelf in the application layout.


ronenrozner:
Thank you for this. Simple but great feature.

xdgc:
@OzzieIsaacs bump, trivial layout change with usability value

PR #3358, opened 2025-04-08
Update kobo.py to include subtitle if field exists in custom columns

dotknott:
Since Subtitle is becoming more common in Calibre DB setups with Kobo users, I added subtitle to the sync data when there's a custom column labeled 'subtitle'. When no matching column exists it returns an empty string.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -426,7 +426,21 @@ def get_series(book):
         return None
     return book.series[0].name
 
-
+def get_subtitle(book):
+    #ID the Column id associated with Subtitle
+    col = (calibre_db.session.query(db.CustomColumns)
+                      .filter(db.CustomColumns.mark_for_delete == 0)
+                      .filter(db.CustomColumns.datatype.notin_(db.cc_exceptions))
+                      .filter(db.CustomColumns.label == 'subtitle') 
+                      .order_by(db.CustomColumns.label).all())[0]
+    
+    if (col):
+        subtitleColumn = getattr(book, 'custom_column_' + str(col.id))
+        if len(subtitleColumn):
+            return subtitleColumn[0].value
+    else:
+            return ""
+        
 def get_seriesindex(book):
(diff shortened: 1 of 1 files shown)


szkafen:
In my particular case, I wanted the subtitle field to show the number of pages as, e.g., "256 pages". This solution seems to only like columns with plain text or numbers. I made the #subtitle column be populated with integers from Count Words and it worked, however it only outputed "256", which was only half my desired end goal.

I tried making a new #subtitle column, having it referencing the number column and with the template {#pages:} pages. The output of the column was indeed "256 pages", but sync would fail allways. It seems it doesn't like templates.

However, I found a solution. Going back to making the #subtitle column be the one populated by Count Pages and changing line 440 to `return str(subtitleColumn[0].value) + " pages"` worked flawlessly, and outputed the desired outcome in the Kobo. Here, " pages" can be changed to the desired word (in my case, since I want to respect Spanish grammar, " páginas"). It could also be changed to `return "Pgs: " + str(subtitleColumn[0].value)` or whatever to show variants of this output.

AsherMaximum:
I've made [#3376](https://github.com/janeczku/calibre-web/pull/3376) to make the subtitle sent to Kobo devices a config option in the Admin Feature settings, along with an option to add a suffix and/or prefix to it, which should cover both @dotknott's and @szkafen's use cases.

new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) `v4.0.99` ([commit `44a32bf`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/44a32bf)) — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.99`. Kobo devices now receive the Subtitle from a Calibre custom column labeled `subtitle`. Rewrote the helper for empty-result safety (no custom column → empty string instead of IndexError).

Thanks @dotknott for the original PR.

Issue #2401, opened 2022-05-08
Kobo sync: allow setting request scheme when behind TLS proxy


ludoo:
I run calibre-web via Docker compose on port 8083, and expose it via Traefik on port 443 with a TLS certificate.

Everything works, except book syncing from Kobo where books are never actually downloaded.

I tried nailing down the issue and it seems to be due to the scheme set in metadata response for `DownloadUrls`, as the actual book download URLs have scheme set to `http` instead of `https`:

```json
"DownloadUrls": [
      {
        "Format": "EPUB3",
        "Size": 499610,
        "Url": "http://calibre-web.xxx.xx.xx/kobo/000000000000000000000000000/download/37/epub",
        "Platform": "Generic"
      },
```

My suspicion is that calibre-web [looks at the request coming from my reverse proxy](https://github.com/janeczku/calibre-web/blob/c1f1952b04cc8bb489dff94d88d451eebdcbc0e7/cps/kobo.py#L375) and sets the scheme based on that.

A simple fix would be inferring the scheme from the port, so if I set `443` as the external port, the scheme is set to `https` regardless.

OzzieIsaacs:
Please use the X-Scheme header as described in the wiki (https://github.com/janeczku/calibre-web/wiki/Setup-Reverse-Proxy#nginx) to archive this:
`proxy_set_header        X-Scheme        $scheme;` 
(or a hardcoded https)
The request scheme in the line you are refering to is set to https here: https://github.com/janeczku/calibre-web/blob/c1f1952b04cc8bb489dff94d88d451eebdcbc0e7/cps/reverseproxy.py#L74

OzzieIsaacs:
Sorry wrong example in the last post. But there is also one for traefik

OzzieIsaacs:
https://github.com/janeczku/calibre-web/wiki/Setup-Reverse-Proxy#traefik-260-example-kobo-sync-fix-enabled

ludoo:
@OzzieIsaacs that's beautiful, thanks a lot for the links! I would add a mention in the wiki page for Kobo, to avoid similar questions in the future.

Thanks a lot!!!

PR #3369, opened 2025-04-19
Allow Overriding Reverse Proxy Configuration with Environment Variables

chtzvt:
Hey there, 

I recently encountered an identical issue to the problem described in https://github.com/janeczku/calibre-web/issues/2401, and needed to implement a workaround that didn't rely on appending headers in the reverse proxy. 

In my case, I use an [opaque reverse proxy](https://tailscale.com/kb/1223/funnel) that isn't configurable, and I realized that others may encounter the same problem depending on their deployment scenario. Thus, this PR introduces an optional mechanism that allows the same set of configuration you'd otherwise apply in a reverse proxy's settings to be set with environment variables where calibre-web itself is hosted. 

I can confirm that this works in my testing, as Kobo download links are generated properly and respect the overridden scheme, host, and port. 

## TL;DR

In addition to overriding request environment settings via `X-Scheme` and similar headers, calibre web will now respect the following overrides via environment variables:

- To configure a path prefix override (script name), set PROXY_SCRIPT_NAME
- To configure a scheme override, set PROXY_SCHEME
- To configure a host override, set PROXY_HOST
- To configure a port override, set PROXY_PORT

### DIFF
--- a/cps/reverseproxy.py
+++ b/cps/reverseproxy.py
@@ -36,6 +36,7 @@
 #
 # Inspired by http://flask.pocoo.org/snippets/35/
 
+import os
 
 class ReverseProxied(object):
     """Wrap the application in this middleware and configure the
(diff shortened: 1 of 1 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) `v4.0.100` ([commit `d088909`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/d088909)) — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.100`. For anyone behind Tailscale Funnel, Cloudflare Tunnel, or another opaque TLS terminator: `PROXY_SCRIPT_NAME` / `PROXY_SCHEME` / `PROXY_HOST` / `PROXY_PORT` now configure the middleware from env. `X-Forwarded-Proto` also recognized alongside `X-Scheme`.

Thanks @chtzvt for the original PR.

PR #3374, opened 2025-04-24
Update deprecated mimetypes for cbr and cbz in __init__.py

mkduk:
This is a patch for issue 3373.

IANA deprecated the mimetypes "application/x-cbr" and "application/x-cbz" in 2017 and replaced them with "application/vnd.comicbook-rar" and "application/vnd.comicbook+zip". Newer readers may not recognize the deprecated mimetypes. This fork changes those types when mimetypes are added in __init__.py and nothing else.

IANA documentation for the deprecation and replacement:
[CBR](https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar)
[CBZ](https://www.iana.org/assignments/media-types/application/vnd.comicbook+zip)

Merging this PR may not be desirable as older readers are still going to look for the old mimetype. But at least one reader (Cantook) won't read CBR/CBZ files without the new mimetype. I understand this probably can't really be "fixed", but I'm just documenting it here so that anyone else wondering why their reader can't see CBR and CBZ files in Calibre-Web's OPDS.


gelbphoenix:
Couldn't we have both for a time being?

OzzieIsaacs:
There had been other people copmplaining like #3421
, I got it fix with the current mimetype, which seems to be aligned with calibre, so I'd like to keep it as it is

PR #3387, opened 2025-05-04
Lazy-load book covers

oskardotglobal:
Title. This PR makes calibre-web feel a thousand times snappier.
In the future, we should probably look into serving the covers at another size and maybe even in different, optimized formats (webp, AV1).


ronenrozner:
Fantastic change! Makes a huge difference, especially on self-hosted instances. Thank you.

OzzieIsaacs:
Calibre-web automatically generates thumbnails if configured

PR #3412, opened 2025-06-07
Added account-wise sync for pdf files.

dshowfred:
User is able to resume from reading same page they left so long as they log in using the same account. I created a extra table to track usage and bookmark progress.


gelbphoenix:
Could you maybe make a new PR with only the relating changes for your feature. I don't think that e.g. the Issue Template files need to be changed for your feature to function.

OzzieIsaacs:
I agree with @gelbphoenix. This is unmergable. And please use in the next PR the "Bookmark" table already present in ub.py. I'm fine with extending the table, there is no need for a duplication of this table

PR #3418, opened 2025-06-12
added opds shelf path sorted by new

Volterxien:
Sorting a shelf by new for doing OPDS sync

### DIFF
--- a/cps/opds.py
+++ b/cps/opds.py
@@ -419,6 +419,45 @@ def feed_shelf(book_id):
                 log.error_or_exception("Settings Database error: {}".format(e))
     return render_xml_template('feed.xml', entries=result, pagination=pagination)
 
+@opds.route("/opds/shelf/<int:book_id>/new")
+@requires_basic_auth_if_no_ano
+def feed_shelf_new(book_id):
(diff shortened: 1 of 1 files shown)


Volterxien:
I'm not sure if it's super viable, but I want this to not paginate and just grab everything. I could create another path but I understand that might not be a super valid use case for most people

Volterxien:
Adding that I no longer need to lack of pagination, so this is fine as is

Volterxien:
@Syer10 Thanks for that!

OzzieIsaacs:
? Don't get it
The endpoint is not linked to the opds tree in any way. To get the content of a shelf ordered accoring to "new" use the webui to sort it accordingly (can be done with every shelf individually).
I don't think the code does work anyhow, as the variable entries is not defined:
grafik

Volterxien:
I didn't test the latest patch, but the reason I wanted this is that OPDS would see the shelf in whatever order it had last been sorted by.  
Also, if sorted by new, it wouldn't refresh the sort order until you reloaded the shelf in the web UI.  
This allows for a consistently ordered OPDS path.

I can retry this with the latest version and get it to work, but if you don't want to add this path, that's fine too! Just trying to share code I'm using.

PR #3429, opened 2025-07-16
add generic oidc authentication method

firlevapz:
Allow generic OIDC authentication agains any OAuth2 provider, re-using the flask_dance library.

- upgrade styling a bit for oauth2 configuration
- add option to automatically create users which are authenticated agains a configured provider
- adapt `reverseproxy.py` to use forwarded protocol when behind reverse proxy

### DIFF
--- a/cps/reverseproxy.py
+++ b/cps/reverseproxy.py
@@ -53,6 +53,18 @@ class ReverseProxied(object):
         proxy_set_header X-Scheme $scheme;
         proxy_set_header X-Script-Name /myprefix;
         }
+
+    In Caddy (modern syntax):
+    reverse_proxy localhost:8083 {
+        header_up X-Forwarded-Proto {scheme}
+        header_up X-Forwarded-Host {host}
+    }
+
+    Or using legacy headers:
+    reverse_proxy localhost:8083 {
+        header_up X-Scheme {scheme}
+        header_up X-Forwarded-Host {host}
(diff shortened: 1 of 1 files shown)


firlevapz:
the old OAuth config is still there, just looks a bit nicer now:

image

this is how the new Oidc Oauth looks like:

image

and here the login form:

image

gelbphoenix:
The maintainer has said multiple times that he doesn't want to add another OIDC provider, so I'm not sure that they will merge this one.

If you want to use a custom OIDC/OAuth provider, then my [Autocaliweb](https://github.com/gelbphoenix/autocaliweb) has already an implementation which also supports `/.well-known/openid-configuration` links.

firlevapz:
ok, let's see, maybe the generic OIDC one will be accepted (as this would the only one to make sense, when you just support a few, because it's generic...)

thanks for the link to your project, will give it a try!

OzzieIsaacs:
No more login routines, sorry, I can't support it

jdwill916:
@OzzieIsaacs You are already supporting this. All this does is open up the existing Oauth support so any provider will work, not just a couple of hardcoded ones.

tedtramonte:
Just to emphasize, this is the exact same login routine that is already implemented. If anything, it could reduce your support burden by removing two hard coded methods. Please reconsider!

PR #3289, opened 2025-01-17
add goodreads metadata provider

claireratigan:
This PR adds a metadata provider for goodreads, taking some inspiration from https://github.com/maria-antoniak/goodreads-scraper

PR #3433, opened 2025-07-24
Add Goodreads metadata provider

jbordalo:
Calibre-web was missing a Goodreads metadata provider.
This one follows the amazon.py one to keep it in the same style.
The author parsing could be simpler, but I kept a less simple version that lists the author's contribution, i.e. "editor"


OzzieIsaacs:
Sorry for the late reply. I will not add any new metadata providers:
https://github.com/janeczku/calibre-web/wiki/Developer's-corner#add-metadata-provider

You could link it in the wiki and "host it on your on github repository
Have also a look at: #3289

PR #3443, opened 2025-08-18
Fix: Setting read status for archived books in the books list page

rjaakke:
Changing the read status in the booklist would fail for archived books with an error the the custom_column doesn't exist.
Unarchived books didn't have this issue.

The issue was that call to retrieve the book by default does not include archived books.
Setting the second parameter in the function call to True ensures that archived books are included.


OzzieIsaacs:
Having a custom column linked to the read status is needed for this bug to occur

PR #3463, opened 2025-10-11
Modernize tooling with uv (ruff, pre-commit); preserve PyPI imports

knguyen1:
# Summary

* This is a…
  * [x] Tooling modernization
  * [x] Bug fix (editable build/discovery)
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change
* Describe this change in 1-2 sentences:
  Adopt the uv toolchain (uv lock/sync, ruff, pre-commit) and adjust packaging so editable installs work locally while keeping the public `calibreweb.*` import path unchanged.

# Problem

* Local editable installs failed (multiple top-level packages; entry points targeting a non-existent `calibreweb.*` in this repo layout).
* Pre-commit pip-compile hooks were placeholders, ran unconditionally, and referenced a non-existent “build” group.
* Resolver friction for Python 3.7 given current dependency pins.

# Solution

Intent: modernize the developer workflow around uv while preserving published interfaces.
- Packaging (uv-friendly, no public API change)
  - Map `calibreweb` → `cps` via `tool.setuptools.package-dir`.
  - Restrict discovery to `calibreweb*`; exclude `library*`.
  - Script: `calibreweb.main:main`.
  - Dynamic version: `calibreweb.constants.STABLE_VERSION`.
  - Keep `MANIFEST.in` as `graft cps` (optionally `prune library`).
  - Set `requires-python = ">=3.8"` to align with dependency support and enable uv resolution.
- Tooling (ruff, pre-commit, uv)
  - Keep ruff hooks; add/normalize uv hooks.
  - Add pip-compile hooks per extra (comics, gdrive, gmail, goodreads, ldap, oauth, metadata, kobo) and for `dev` group.
  - Target Python 3.8 (lowest supported) for deterministic cross-version constraints.
  - Run pip-compile only when `pyproject.toml` or `uv.lock` change.
  - Remove unused “build” pip-compile hook.

# Action

Additional actions required:
* [ ] None

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,3 +44,86 @@ In case your code enhances features of Calibre-Web: Create your pull request for
 Please check if your code runs with python 3, python 2 is no longer supported. If possible and the feature is related to operating system functions, try to check it on Windows and Linux.
 Calibre-Web is automatically tested on Linux in combination with python 3.8. The code for testing is in a [separate repo](https://github.com/OzzieIsaacs/calibre-web-test) on GitHub. It uses unit tests and performs real system tests with selenium; it would be great if you could consider also writing some tests.
 A static code analysis is done by Codacy, but it's partly broken and doesn't run automatically. You could check your code with ESLint before contributing, a configuration file can be found in the projects root folder.
+
+
+### Developer Workflow
+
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
This breaks my build chain and has no benefit for me

PR #3464, opened 2025-10-12
Update messages.po

Trashwarez:
Added missing German translations and made some corrections


OzzieIsaacs:
Sorry I picked the other german translation instead of your. 

Trotzdem Danke für die Mühe

PR #3470, opened 2025-10-22
Added missing german translations and fixed some incorrect ones

holzerseb:
I've added the missing german translations and fixed a few incorrect translations. I did NOT check the entire file, just fixed what catched my eye during scrolling.

Please also note, the split apart localization strings in line 3120 and below ("Book will be deleted from database" and the "and the hard drive") don't work in german. The "deleted" will always be added at the end in a proper german translation. The translation I wrote theoretically works from a grammatical perspective, but is really not how any german speaker would ever say or write it.
Please consider using two separate localization strings for both options ("Book will be deleted from database" and "Book will be deleted from database and hard disk") instead, giving the flexibility across all languages for a proper translation. The way it was originally may result in the translation to be "Book will be from database" if the hard disk string isn't appended, which is why I changed it to what sounds weird but is acceptable: "Book will be deleted, from database (and hard disk)".


OzzieIsaacs:
Thanks. 
Regarding the one string. Yes you are right. 

Ich weiß nicht was mich da geritten hat, ich hab's geändert

PR #3478, opened 2025-11-02
use proxied scheme when building urls

rrgeorge:
When running behind a reverse proxy with SSL, the request.scheme is set from the proxy, not the original request. So this uses the X-Forwarded-Proto header for the scheme if its present.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -358,9 +358,12 @@ def get_download_url_for_book(book_id, book_format):
             host = "".join(request.host.split(':')[:-1])
         else:
             host = request.host
-
+        if "X-Forwarded-Proto" in request.headers:
+            scheme = request.headers["X-Forwarded-Proto"]
+        else:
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
The path you applied the change to is only entered if the request is not proxied, furthermore, the headername is changed to the WSGI servers to: HTTP_X_FORWARDED_PROTO (HTTP_X in the beginning an all in capital letters).
So this patch is useless, or I don't understand the sense of it

rrgeorge:
The proxy this is referring to is a reverse proxy, such as nginx or caddy, sitting in front of calibre-web so that you can have an SSL site as described here: https://github.com/janeczku/calibre-web/wiki/Configuration#reverse-proxy-for-secure-self-hosting
The problem is that nginx communicates with calibre-web via http, so the request.proto in this case is only http. This patch would use the header as provided by the reverse proxy server if it is set.

PR #3491, opened 2025-11-27
Fix display issue on small width resolution and rename Google metadata provider

diginfotek:
**1. Caliblur theme display fix**
With the caliblur theme, there is a display issue when having a small resolution width (e.g. split screen, etc.)
Screenshot 2025-11-27 121627

This also happens if using extra providers
Screenshot 2025-11-27 134341

The fix moves the yellow message to the top, like the standard theme. The other option would have been to replicate the mobile layout, e.g. hiding the message.
Screenshot 2025-11-27 121711

**2. Google Books metadata provider**
Renamed Google metadata provider to Google Books.

Currently, the Google Books metadata is named Google, which could be confused with the search engine being used as a metadata provider.  Replicating the naming of Google Scholar by renaming Google to Google Books, makes the list of metadata providers clearer.

### DIFF
--- a/cps/metadata_provider/google.py
+++ b/cps/metadata_provider/google.py
@@ -31,7 +31,7 @@ log = logger.create()
 
 
 class Google(Metadata):
-    __name__ = "Google"
+    __name__ = "Google Books"
     __id__ = "google"
     DESCRIPTION = "Google Books"
     META_URL = "https://books.google.com/"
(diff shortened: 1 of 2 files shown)


diginfotek:
Please let me know if it's better to split from master into a new branch, as I originally worked on it before replicating on the Develop branch

Issue #1998, opened 2021-05-26
Google Oauth error


Kuroburn:
**Describe the bug/problem**
I'm getting Internal Server Error 500 when trying to use the google icon in the login screen. I already did all the steps in the wiki and read all possible mistakes I could do but nothing seems to be my case.

**To Reproduce**
1. I installed #Gdrive and Gmail integration, #Gmail, #oauth and #Kobo integration from optional-requirements.txt 
2. I created the project in Google Cloud Platform and did the steps from wiki to setup the enviroment
3. I generated the credentials, downloaded the .json file and put everything inside  the calibre-web folder (client_secrets.json) and webapp (credentials).
4. I setup the environment variable in a **_.sh_** file inside _/etc/profile.d/_ since when rebooting the server the variable gets unseted (this is a point to emphasize on the wiki, at least from CentOS)
5. I restarted the app and rebooted the server and... when try to login... error 500.

**Logfile**
<details>
  <summary>Click here</summary>
[2021-05-26 01:22:47,766]  INFO {cps:101} Starting Calibre Web...
[2021-05-26 01:22:47,954]  INFO {cps.server:227} Starting Tornado server on :443
[2021-05-26 01:23:23,258] DEBUG {cps.web:1570} User logged out
[2021-05-26 01:23:30,825] ERROR {cps:1455} Exception on /login/google/authorized [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 2070, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1515, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1513, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1499, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/local/lib/python3.6/site-packages/flask_dance/consumer/oauth2.py", line 267, in authorized
    **self.token_url_params,
  File "/usr/local/lib/python3.6/site-packages/requests_oauthlib/oauth2_session.py", line 360, in fetch_token
    self._client.parse_request_body_response(r.text, scope=self.scope)
[...]

PR #3493, opened 2025-12-03
Update Google OAuth scope to include 'openid' 

um-jglad:
Google includes `openid` by default, so configuring Google oAuth breaks unless you use the environment variable `OAUTHLIB_RELAX_TOKEN_SCOPE` as listed in the wiki. Seems like it would be better to just patch this, but I may be misunderstanding the root issue.

Tested on Python 3.11.2

Related Issue: #1998


um-jglad:
I'm pretty sure this was just me misunderstanding the issue I was facing. OAUTHLIB_RELAX_TOKEN_SCOPE is still required and the patch doesn't make a difference it seems.

PR #3504, opened 2025-12-20
Chore: Use natsort in cps/comic.py (when available)

lb803:
Improve comic page sorting using natsort.

Replace standard `sorted()` with natsort (when available) to better handle comic page filenames when ordering pages to select cover images.

### DIFF
--- a/cps/comic.py
+++ b/cps/comic.py
@@ -30,6 +30,13 @@ except (ImportError, RuntimeError) as e:
 log = logger.create()
 
 try:
+    from natsort import natsorted as sort
+except ImportError:
+    log.debug("Natural sorting unavailable: using standard "
+              "sorted() method instead.")
+    sort = sorted
+
+try:
     from comicapi.comicarchive import ComicArchive, MetaDataStyle
     use_comic_meta = True
     try:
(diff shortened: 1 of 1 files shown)


lb803:
**The problem**
To create a book cover for comic ebooks, Calibre-web gets an ordered list of the files contained in the e-book and assumes that the first one is the cover page. This works for the great majority of cases, but when filenames are named improperly, this fails.

**The fix**
Use a better function to get an ordered list of the filenames, specifically using a natural language sorting approach.

**The implementation**
Calibre-web codebase already makes use of the natsort function; this PR simply reuses it by adopting the same import and use case conventions as in other parts of the codebase.


@OzzieIsaacs if anything looks off or needs adjustment, please let me know. I'm happy to refine the PR.

new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) `v4.0.98` ([commit `db677b1c`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/db677b1c)) — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.98`. For anyone who's been getting random interior pages as their comic covers, the natural-sort lands across all four CBZ/CBT/CBR/CB7 paths.

Thanks @lb803 for the original patch.

PR #3475, opened 2025-10-30
Improve EPUB reader: theme enhance, theme picker, font controls, mobile swipe gestures option, autosave position, etc, & UI polish

IgorKurkov:
Hi guys @janeczku @OzzieIsaacs ! 

I created few commits with different improvements mostly for mobile reading.

Improve the EPUB reader experience with persistent reader settings, UX improvements for mobile/gestures, new theme/color support, and several UI fixes.
Persisted settings (theme, font size) and reading position to localStorage so books open where the user left off.
Added mobile-friendly gestures prev/next, also swipe to close sidebar, (Hammer.js option), new UI items (books back button, pages count), and small bugfixes and polish.

### What changed

**Added**
- Autosave current reading position for each book (opens book at last read point, saved to LS).
 - Custom theme selection with color picker (Pickr) for reader theming (+ option in settings, +saving setting to LS).
- Font size selector with fine grained  + / - controls (+ option in settings, +saving setting to LS).
- Font family selector saved to localStorage.
- Books button, page count display aka 1/1234 (of locations) on bottom, swipe-to-close sidebar feature.
- Optional gesture support via Hammer.js for mobile (swipe controls) (+ option in settings, +saving setting to LS).

**Changed**
- Save/read theme, font size and family from/to localStorage so settings persist.
- UI small polish and styling fixes for reader and settings modal (i.e. use dvh for mobile dynamic height to achieve full height of the screen).

**Fixed**
- Various small reader layout/style issues (including a JS reader height adjustment).
- Allow console process interruption with Ctrl+C (developer convenience).

NB: Each commit has specific feature. I suggest to go through each commit one by one.

**Usage**
- Open the reader and verify theme selection persists after reload.
- Change font size and family and confirm persistence.
- Read a book, navigate to a page, close and reopen the same book — it should open at the saved position.
- On mobile, test swipe gestures and swipe-to-close sidebar.
- Verify page count appears at bottom and in settings.
- Confirm console interrupt works (for any background process you use that this change affects).

PR #3513, opened 2026-01-02
Epub reader adjustments: add Columns setting saving to localStorage (as per asking from comment)

IgorKurkov:
Hi All

as an addition to this PR - https://github.com/janeczku/calibre-web/pull/3475 

I have added desired saving to localstorage for Columns one/two for user [ronenrozner](https://github.com/ronenrozner)

please take a look, nothing special here, just handler and saving

and Happy New Year Everyone :)

### DIFF
--- a/cps/static/js/reading/epub.js
+++ b/cps/static/js/reading/epub.js
@@ -284,5 +284,11 @@ var reader;
         if (savedFont && window.selectFont) {
             window.selectFont(savedFont);
         }
+
+        // Restore saved spread/columns setting
+        const savedSpread = localStorage.getItem("calibre.reader.spread");
+        if (savedSpread && (savedSpread === "spread" || savedSpread === "nonespread")) {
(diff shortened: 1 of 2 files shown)


IgorKurkov:
the problem I encountered, a lot of the code for features, especially **gestures** has been removed (probably accidentally) in commit Merge Reader improvements - https://github.com/janeczku/calibre-web/commit/eb872865a5bc28eaf0080d62b66c9ccf0197c557

Also I can't find **hammer.min.js** in next commits where it has been removed, because it is required for gestures. Did you guys removed it for some reason @OzzieIsaacs ? current code with pointstart pointend didnt work, thats why I added it, current code in master is broken anyway so we need to take a look into failed merge.

it cause current conflict as far as it was happen because of vanishing current code. WIP, suggest to revert this commit, but this is a merge commit... so this is sort of problem

PR #3476, opened 2025-11-01
Optimize search performance for large libraries

alcibiadesc:
## Summary
This PR introduces significant performance improvements to Calibre-Web's search functionality, particularly beneficial for large libraries. Search times are reduced from 3-9 seconds to under 330ms in most cases, representing a 89-97% performance improvement.

## Problem Statement
Users with large Calibre libraries (100,000+ books) experience slow search operations due to:
- Multiple expensive `query.count()` calls generating deeply nested SQL
- N+1 query problems from lazy loading
- Inefficient `.any()` filters with OR conditions across multiple relationships
- Nested loops performing individual database queries for author ordering

### 1. FTS5 Full-Text Search Integration
- **What**: Adds support for SQLite's FTS5 full-text search indexes
- **Why**: FTS5 provides indexed full-text search, dramatically faster than LIKE queries
- **Implementation**: Attempts FTS5 search first, falls back to traditional search if unavailable
- **Benefit**: Sub-second indexed searches with zero breaking changes

### 2. Optimized Subqueries
- **What**: Replaces `.any()` subqueries with JOIN-based subqueries
- **Why**: `.any()` generates expensive EXISTS clauses; JOINs are more efficient
- **Implementation**: Uses `Books.id.in_(subquery)` pattern for tags, series, authors, publishers
- **Benefit**: Simpler SQL, better query planning, faster execution

### 3. Eager Loading with selectinload()
- **What**: Preloads author relationships using SQLAlchemy's selectinload
- **Why**: Prevents N+1 queries when accessing book.authors
- **Implementation**: Added `.options(selectinload(Books.authors))` to base query
- **Benefit**: Single batch query instead of one query per book

### 4. LIMIT+1 Estimation Pattern
- **What**: Uses LIMIT+1 to estimate result counts instead of COUNT(*)
- **Why**: COUNT(*) on large result sets is expensive
- **Implementation**: Fetches limit+1 results, checks if more exist
- **Benefit**: Fast pagination without counting all results

### 5. Dictionary-Based Author Ordering
- **What**: Replaces nested database queries with dictionary lookups
- **Why**: Original code made N queries inside nested loops (O(n²))
- **Implementation**: Builds author dictionaries for O(1) lookups
- **Benefit**: Reduces complexity from O(n²) to O(n)

## Performance Results

Testing environment:
- Library size: 129,141 books, 41,552 authors
- Database size: 224.8 MB
- Platform: Production server with typical workload

| Search Term | Type     | Before  | After   | Improvement |
[...]

Issue #718, opened 2018-12-02
Advanced search with "or" instead of "and"


lionelrousvoal:
Hello

Is there any possibility to launch an advanced search with "or" on different fields?

Regards

Lionel

OzzieIsaacs:
Currently everything is put together using "and", if you qould change this the UI has to be more complicated and less intuitively.

lionelrousvoal:
Sure...

PR #3531, opened 2026-01-24
Full Text Search Has Landed

### LINKED ISSUE #1412: Advanced search is slow
**Describe the bug/problem**
When I type 'heinlein' in the search box and press [Search], I get my results in a second or so.
When I use advanced search and type 'heinlein' in the 'Author' box (and no other search items) and press [Search], it takes 20 - 30 seconds to get results.

**Environment (please complete the following information):**
 - OS: Fedora 31
 - Python version 3.7.7
 - Calibre-Web version master@2020-05-20, 13:30
 - Special Hardware: Intel® Core™ i7-8700 CPU @ 3.20GHz; 32GB
 - Browser: Firefox 76.0

### LINKED ISSUE #2308: Fuzzy search like sublime text
https://www.forrestthewoods.com/blog/reverse_engineering_sublime_texts_fuzzy_match/

**Additional context**
Sometimes we are looking for the name of an author who may have an error or even forget how to write correctly, for example, schopenhauer. If you miss just one letter, you won't find the author.

### LINKED ISSUE #2490: Full text search
I know that in the past when requests were given for full text search you said you had decided not to move in that direction. But given the new version of Calibre 6, with its full-text search, I wonder if things might also change for calibre-web.

I am unable to find books based on searches for things in the text itself. Right now I can only search the metadata, which is less useful than if I could search both the metadata and the full text.

When searching the library, there should be an option to search the full text.

Using Calibre itself, but that doesn't give us the beautiful UI of calibre-web nor the ability to use it online through any browser

**Additional context**
It would be good to still be able to limit searches with metadata (e.g., find any book with the words "I will protect even those I hate so long as it is right" but only return results from books tagged "fantasy fiction")

Thanks!

### LINKED ISSUE #2776: Full Text Content Search Feature
Is the Full Text Content Search Feature is available in web interface? if not, is it possible to add 

Ex:
https://download.calibre-ebook.com/images/4/fts-matches-thumb.jpg

### LINKED ISSUE #3129: search the description with quick search
I have added further information to the description field of many books.
This field can be searched in the advanced search, but not with the quick search.

Would it be possible to include the description as a search source in the quick search?

If so, perhaps with a switch in the options so that only those who need it can activate it?

### LINKED ISSUE #3468: Calibre-Web Search Optimization
## 📋 Summary

This document describes a **radical optimization of Calibre-Web’s search system**, that I implemented, which reduced search times from **9 seconds to 92–329ms**, achieving a **90–97% improvement**.

**Library size**: 129,141 books, 41,552 authors, database size 224.8 MB

---

## 🎯 Initial Problem

Search operations in Calibre-Web originally took between 7–13 seconds due to multiple issues:

1. **`query.count()` generated massive SQL queries** with deeply nested EXISTS clauses  
2. **N+1 problem**: Lazy loading triggered thousands of individual queries  
3. **`generate_linked_query()`** produced extremely complex JOINs incompatible with eager loading  
4. **SQL logging enabled** (`echo=True`) added 3–5 seconds of overhead  
5. **Expensive OR filters** using `.any()` across several relationships (tags, series, authors, publishers)

---

### 1. Removed `query.count()` ✅

**Problem:** `count()` generated huge nested queries, taking over 6 seconds  
**Solution:** Implemented the LIMIT+1 pattern to estimate results:
```python
result = query.limit(61).all()
has_more = len(result) > 60
if has_more:
    result = result[:60]
result_count = len(result)
```
**Result:** Query phase reduced from 6 s → 20 ms

---

### 2. Disabled SQL Logging ✅

```python
engine = create_engine('sqlite://',
                      echo=False,  # Previously echo=True
                      isolation_level="SERIALIZABLE",
                      ...)
```
**Result:** 3–5 s saved per query

---

### 3. Full Bypass of `generate_linked_query()` ✅

Replaced with a simplified query using selective eager loading:
```python
from sqlalchemy.orm import selectinload

base_query = self.session.query(Books).filter(self.common_filters(True))
query = base_query.options(selectinload(Books.authors))
```
**Result:** Loading time reduced from 6 s → ~50 ms

---

### 4. FTS5 as the Primary Filter ✅

Used FTS5 for lightning-fast index lookups:
```python
fts_results = self.session.execute(
    text("SELECT DISTINCT rowid FROM books_fts WHERE books_fts MATCH :term"),
    {"term": term}
).fetchall()
fts_ids = [r[0] for r in fts_results]

if fts_ids:
    query = base_query.filter(Books.id.in_(fts_ids))
else:
    ...
```
**Result:** FTS searches < 2 ms for index, 20–51 ms total

---

### 5. Optimized Fallback with Direct JOIN ✅

Replaced slow `.any()` subqueries with a direct join:
```python
author_subquery = self.session.query(books_authors_link.c.book).join(
    Authors, books_authors_link.c.author == Authors.id
).filter(func.lower(Authors.name).like("%" + term.lower() + "%"))

query = base_query.filter(Books.id.in_(author_subquery))
```
**Result:** Fallback reduced from 833 ms → 220 ms (74% faster)

---

### 6. Batch Loading for Tags/Series ✅

Loaded tags and series **after** the main query for efficiency:
```python
if result:
    book_ids = [b.id for b in result]

    tags_stmt = select(books_tags_link).where(books_tags_link.c.book.in_(book_ids))
    tags_links = self.session.execute(tags_stmt).fetchall()

    series_stmt = select(books_series_link).where(books_series_link.c.book.in_(book_ids))
    series_links = self.session.execute(series_stmt).fetchall()
```
**Result:** Tags/series loading time 4–14 ms

---

### 7. Optimized Author Ordering ✅

Replaced nested loops (O(n²)) with dictionary lookups (O(n)):
```python
authors_dict = {a.sort: a for a in book.authors}
for auth in sort_authors:
    if auth in authors_dict:
        book.ordered_authors.append(authors_dict[auth])
```
**Result:** Sorting reduced from 636 ms → 0–3 ms

---

### 8. Optimized Status Queries ✅

Fetched only the required fields:
```python
archived = ub.session.query(ub.ArchivedBook.book_id).filter(...).all()
archived_set = {a[0] for a in archived}
```
**Result:** Status queries < 2 ms

---

### 9. BookWrapper with `__slots__` ✅

Optimized memory footprint:
```python
class BookWrapper:
    __slots__ = ('Books', 'is_archived', 'read_status', 'id')
    def __init__(self, book, is_archived, read_status):
        self.Books = book
        self.is_archived = is_archived
        self.read_status = read_status
        self.id = book.id
```
**Result:** Memory usage per object reduced

---

## 📊 Final Results

| Term | Type | Before | After | Improvement |
|------|------|---------|--------|-------------|
| futbol | FTS | 3000ms | 92–329ms | 89–97% |
| coche | FTS | 3000ms | 185ms | 94% |
| javascript | FTS | 3000ms | 92ms | 97% |
| ikea | FTS | 3000ms | 85ms | 97% |
| test999 | Fallback | 3000ms | 288ms | 90% |
| xyz123 | Fallback | 3000ms | 290ms | 90% |

**Average search time reduced by 95%**  
**User experience:** Near-instant searches (<330ms vs 3000ms)  
**Server load:** 90% CPU time reduction per query

---

webysther:
> Do you think this PR help you? Consider a donation: https://github.com/sponsors/webysther

Full Text Search is finally here, bringing a faster, more powerful way to discover books. Search now taps into Calibre’s full‑text index with an optional toggle in settings, while preserving the existing search as a safe fallback. Under the hood, the feature uses a dedicated FTS database, loads Calibre’s tokenizer extensions, and integrates cleanly with the existing /search flow. The result is an exciting, revitalizing upgrade to search—more accurate, more responsive, and ready for deeper discovery.

- Allow enable/disable full text search
- Allow merge the simple search to full text search results
- Allow hide/show the archived books on full text search results
- Button to show the search result by relevance
- Bugfix: if search paginate show the total books found, not the current page size (now only fixed for full text search)

image

image

image

https://github.com/user-attachments/assets/e1af129a-7421-4603-89b4-b580eb60ab54

## Current limitations

- Only works on linux for now: https://github.com/janeczku/calibre-web/pull/3531/changes#diff-3d1e6cbcc36eb1e2f55fa003061e5c72ad54aa8fff544fc201d1d326ef16c5cdR452
- Necessary to set 2 new ENV on LinuxServer container:

```yaml
LD_LIBRARY_PATH=/app/calibre/lib
DOCKER_MODS=linuxserver/mods:universal-calibre
```

fix https://github.com/janeczku/calibre-web/issues/2490 
fix https://github.com/janeczku/calibre-web/issues/2776 
fix https://github.com/janeczku/calibre-web/issues/3129 
fix https://github.com/janeczku/calibre-web/issues/3468 
fix https://github.com/janeczku/calibre-web/issues/2308 
fix https://github.com/janeczku/calibre-web/issues/1412

related https://github.com/janeczku/calibre-web/pull/3476 
related https://github.com/janeczku/calibre-web/pull/2772 
related https://github.com/janeczku/calibre-web/issues/718


crocodilestick:
For those interested, this functionality has now been added to Calibre-Web Automated (CWA) in addition to a full management console & automated indexing for the process that allows users to better control over the process as well as the ability to view the current indexing status of their library ect. Amazing work @webysther <3

crocodilestick:
Hey @webysther - I had to pull the FTS changes from the upcoming release. The latest version still has a few blocking issues that would ship broken behaviour:

The FTS SQL orders by rank but never selects a rank field, so the query errors and FTS silently falls back. That means relevance never actually works.
The FTS query dropped common_filters(True), which can expose books that users shouldn’t see (archived/restricted).
The merge logic combines already‑paginated base results with full FTS results, then paginates again. That makes counts and paging wrong.
Because of those, I reverted it from main and kept it on a separate full-text-search branch/PR so it can be finished safely.

What still needs fixing before it’s ready:

Restore a valid FTS ranking query (or select a rank column) so relevance sorting actually works.
Re‑apply common_filters(True) to keep permissions/visibility correct.
Fix pagination/counts by paginating after merging full result sets, or fetch unpaginated base results when merging.
Verify extension loading is correct on Python/SQLite (enable load extension before load_extension).

Once those are fixed and tested, I’ll bring it back into CWA!

webysther:
> Hey @webysther - I had to pull the FTS changes from the upcoming release. The latest version still has a few blocking issues that would ship broken behaviour:
> 
> The FTS SQL orders by rank but never selects a rank field, so the query errors and FTS silently falls back. That means relevance never actually works. The FTS query dropped common_filters(True), which can expose books that users shouldn’t see (archived/restricted). The merge logic combines already‑paginated base results with full FTS results, then paginates again. That makes counts and paging wrong. Because of those, I reverted it from main and kept it on a separate full-text-search branch/PR so it can be finished safely.
> 
> What still needs fixing before it’s ready:
> 
> Restore a valid FTS ranking query (or select a rank column) so relevance sorting actually works. Re‑apply common_filters(True) to keep permissions/visibility correct. Fix pagination/counts by paginating after merging full result sets, or fetch unpaginated base results when merging. Verify extension loading is correct on Python/SQLite (enable load extension before load_extension).
> 
> Once those are fixed and tested, I’ll bring it back into CWA!

All these are fixed already!

Astromormy:
Has this change been added yet?

clawoflight:
Can we get this merged?

Astromormy:
There was a release just yesterday, but this merge wasn't included. Why is that? Is something broken? Releases for this project happen once or twice a year, and given how late it is in the current year already, there's no guarantee there'll be another in this one. Are we really going to need to wait until next year to get this feature merged in? It feels frustrating....

Issue #3539, opened 2026-01-30
no such column: books.isbn


mcgovern571:
After upgrading to Calibre 9.0, when starting Calibre Web it immediately fails with the below error.
I have re-installed Calibre-web, I have deleted the app.db, but both immedately result in the same error.
Calibre itself opens with no issue and reports no problems with the database.

Calibre-Web
500 Internal Server Error
The server encountered an internal error and was unable to complete your request. There is an error in the application.
Traceback (most recent call last):
File "sqlalchemy\engine\base.py", line 1967, in _exec_single_context
File "sqlalchemy\engine\default.py", line 941, in do_execute
sqlite3.OperationalError: no such column: books.isbn
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
File "flask\app.py", line 1473, in wsgi_app
File "flask\app.py", line 882, in full_dispatch_request
File "flask\app.py", line 880, in full_dispatch_request
File "flask\app.py", line 865, in dispatch_request
File "cps\usermanagement.py", line 97, in decorated_view
File "cps\cw_login\utils.py", line 296, in decorated_view
File "cps\web.py", line 812, in index
File "cps\web.py", line 418, in render_books_list
File "cps\db.py", line 852, in fill_indexpage
File "cps\db.py", line 862, in fill_indexpage_with_archived_books
File "sqlalchemy\orm\query.py", line 2673, in all
File "sqlalchemy\orm\query.py", line 2827, in _iter
File "sqlalchemy\orm\session.py", line 2362, in execute
File "sqlalchemy\orm\session.py", line 2247, in _execute_internal
File "sqlalchemy\orm\context.py", line 305, in orm_execute_statement
File "sqlalchemy\engine\base.py", line 1418, in execute
File "sqlalchemy\sql\elements.py", line 515, in _execute_on_connection
File "sqlalchemy\engine\base.py", line 1640, in _execute_clauseelement
File "sqlalchemy\engine\base.py", line 1846, in _execute_context
File "sqlalchemy\engine\base.py", line 1986, in _exec_single_context
File "sqlalchemy\engine\base.py", line 2355, in _handle_dbapi_exception
File "sqlalchemy\engine\base.py", line 1967, in _exec_single_context
File "sqlalchemy\engine\default.py", line 941, in do_execute
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: books.isbn
[...]

Issue #3538, opened 2026-01-30
Calibre-web UI won't load after update (Unraid)


PearsonFlyer:
**Describe the bug/problem** 

This morning on Unraid, I updated the Calibre-web container, and now when I load the UI, I get this error:

E-Books
500 Internal Server Error
The server encountered an internal error and was unable to complete your request. There is an error in the application.
Traceback (most recent call last):
File "/lsiopy/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
self.dialect.do_execute(
File "/lsiopy/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 952, in do_execute
cursor.execute(statement, parameters)
sqlite3.OperationalError: no such column: books.isbn
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 1511, in wsgi_app
response = self.full_dispatch_request()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 919, in full_dispatch_request
rv = self.handle_user_exception(e)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
rv = self.dispatch_request()
^^^^^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 902, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/calibre-web/cps/usermanagement.py", line 97, in decorated_view
return login_required(func)(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/calibre-web/cps/cw_login/utils.py", line 296, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/calibre-web/cps/web.py", line 812, in index
return render_books_list("newest", sort_param, 1, page)
[...]

Issue #3537, opened 2026-01-30
Server error after updating calibre desktop version to 9.0. (Mac OS)


AndreasLilies:
Receiving an internal server error after updating my MacOS desktop version of calibre to 9.0 and importing a new ebook. When clicking on unread, read books or even the 'about' menu entry on the left side of the screen, I get similiar error messages like the one below. I'm using version 0.6.25-ls367 Build-date:- 2026-01-25T02:45:22+00:00, all other components (like Python) are not visible as the 'about' menu crashes as well. Tried to restart, reconnect etc. w/o success? Any ideas?

Thx
Andreas

Traceback (most recent call last):
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 1511, in wsgi_app
response = self.full_dispatch_request()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 919, in full_dispatch_request
rv = self.handle_user_exception(e)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
rv = self.dispatch_request()
^^^^^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.12/site-packages/flask/app.py", line 902, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/calibre-web/cps/usermanagement.py", line 97, in decorated_view
return login_required(func)(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/calibre-web/cps/cw_login/utils.py", line 296, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/calibre-web/cps/web.py", line 821, in books_list
return render_books_list(data, sort_param, book_id, page)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3540, opened 2026-01-30
🐛 Calibre 9 compatibility

webysther:
> Do you think this PR help you? Consider a donation: https://github.com/sponsors/webysther
 
fix https://github.com/janeczku/calibre-web/issues/3537 
fix https://github.com/janeczku/calibre-web/issues/3538 
fix https://github.com/janeczku/calibre-web/issues/3539

Tested on Calibre 8.16.2 and Calibre 9.0.0

Removed unused columns while maintaining compatibility with older Calibre versions, as they were moved the isbn data to the table identifiers years ago.

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -416,8 +416,16 @@ class Books(Base):
     path = Column(String, default="", nullable=False)
     has_cover = Column(Integer, default=0)
     uuid = Column(String)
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
Don't make things so complicated. Both columns are not used, so I deleted it

PR #3555, opened 2026-02-03
Hide read books from Discover. Resolves #2588

### LINKED ISSUE #2588: Unread filter for Discover Books/Custom Filters per Page
I'm using the "Discover" page to find new reads for myself, but i didn't find a way to filter out any read books so its full of already read books.

I would like to add a setting that allows to filter the discover page based on the read column, but i'm not sure if that would be the best option. Maybe it would be better if you could define cusom filters per page like you already can on a user basis(Allow/Deny)

**Additional context**
I would like to implement this feature myself but i would like some input on what would be the better aproach for my solution. Maybe somone else has a better idea which would also solve my problem.

lpinner:
Hide read books from Discover. Resolves #2588.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -1843,15 +1843,14 @@ def _configuration_update_helper():
             _config_string(to_save, "config_upload_formats")
 
         _config_string(to_save, "config_calibre")
-        _config_string(to_save, "config_binariesdir")
         _config_string(to_save, "config_kepubifypath")
-        if "config_kepubifypath" in to_save:
-            if config.config_kepubifypath:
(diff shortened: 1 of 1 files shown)


new-usemame:
For anyone landing here looking for a build that has this — backported cleanly into the community-maintained CWA continuation as v4.0.37: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.37

Pull: `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.37`

Thanks @lpinner — your patch applied unmodified, the coalesce on the read-status filter is the load-bearing detail (without it, books a user has never opened would NULL-compare and drop out of the result set entirely).

PR #3591, opened 2026-02-20
feature: allow fetching books from URL directly

Krawabbel:
added ability to directly fetch books from URL, see video

[Demo](https://github.com/user-attachments/assets/c709dd7e-40b8-4a0d-b623-d66fd6bdbb23)

# Background

I frequently add books to my library that can be downloaded from the internet, e.g. from Project Gutenberg. Instead of first downloading them to my local computer or phone and then uploading them again, it seems more convenient to enable the server to directly download the resource. I checked both the issues and open PRs. None of them seem to address this "feature".

# Details

* This is my first-ever contribution to open-source software. I'm so proud that I finally managed to find something that's probably worth a contribution! Of course, I checked the contribution guidelines but, please, be gentle if I missed anything. :)
* I did use the assistance of AI for the frontend changes since I know very little about HTML, CSS and Javascript. However, the feature is far from being vibe-coded. Of course, I vetted all the AI suggestions to the best of my ability but it is very possible that copilot messed something up and I don't see it. The backend code is 100% written by a human (with the help of StackOverflow).
* I'm also not a UI/UX designer. The new dropdown menu could be much nicer but this is the best I (and the copilot) could do.
* I don't know much about cyber security but it is possible that I have created a severe vulnerability with this feature - even though I am not sure how it would be any more severe than allowing file uploads. I made it so that the downloaded file is still passed through the same backend pipeline as if I simply uploaded the file directly.
 
# Auxiliary Contributions

Appended contribution guidelines with guide how to update translation files and updated translation files.


Krawabbel:
Hey @OzzieIsaacs, the PR should be ready for review. Feel free to have a look at it whenever it suits you 😊

PR #3593, opened 2026-02-21
Fix AttributeError on unauthenticated OPDS requests

### LINKED ISSUE #3592: OPDS endpoint returns 500 when accessed without Authorization header
## Bug Description

Accessing `/opds` without an `Authorization` header results in a `500 Internal Server Error` instead of a `401 Unauthorized`. The crash occurs in flask-limiter's `before_request` handler, which calls `request_username()` as its `key_func` before the route's auth decorator has a chance to run. When no credentials are present, `request.authorization` is `None`.

Narrow scope: standard OPDS clients (Kybook, Moon+ Reader, etc.) always send Basic Auth headers, so they are unaffected. Reproduces in setups without anonymous access enabled where the OPDS URL is hit without credentials.
## Steps to Reproduce

1. Disable anonymous access in calibre-web settings
2. `curl http://<host>:8083/opds` (no credentials)
3. Observe 500
## Log File

File "/usr/lib/python3/site-packages/flask/app.py", in preprocess_request rv = self.ensure_sync(before_func)()
File "/usr/lib/python3/site-packages/flask_limiter/_extension.py", in _check_request_limit raise e
File "/usr/lib/python3/site-packages/flask_limiter/_extension.py", in __evaluate_limits limit_key = lim.key_func()
File "/usr/lib/calibre-web/cps/main.py", line 27, in request_username return request.authorization.username
AttributeError: 'NoneType' object has no attribute 'username'
## Expected Behavior
A `401 Unauthorized` response, consistent with other unauthenticated requests.
## Screenshots
N/A
## Environment

- OS: Arch Linux
- Python: 3.14
- Calibre-Web: 0.6.26
- flask-limiter: 4.1.1 (within declared `>=2.3.0,<4.2.0`)
- No Docker, no reverse proxy
## Additional Context
I intend to follow up with a PR with the fix.

RafikFarhad:
## Problem

`/opds` returns `500 Internal Server Error` instead of `401 Unauthorized` when accessed without an `Authorization` header.

`request_username()` is registered as flask-limiter's `key_func` for the OPDS blueprint. Flask-limiter evaluates `key_func` inside a `before_request` handler, which runs before the route's `@requires_basic_auth_if_no_ano` decorator. When no credentials are supplied, `request.authorization` is `None`:

```python
def request_username():
    return request.authorization.username  # AttributeError when no auth header
```

`swallow_errors=True` on the `Limiter` instance does not help — it only suppresses storage backend errors, not exceptions raised inside `key_func`.

## Fix

Guard against `None` so unauthenticated requests use an empty string as the rate-limit key, and the auth decorator handles the `401` correctly:

```python
def request_username():
    return request.authorization.username if request.authorization else ""
```

## Impact

Narrow scope. Standard OPDS clients always send Basic Auth headers and are unaffected. The 500 surfaces only when the OPDS endpoint is hit without credentials (e.g. browser access, local setup without anonymous access enabled, misconfigured client).

Closes #3592

### DIFF
--- a/cps/main.py
+++ b/cps/main.py
@@ -24,7 +24,7 @@ from flask import request
 
 
 def request_username():
-    return request.authorization.username
+    return request.authorization.username if request.authorization else ""
 
 
 def main():


OzzieIsaacs:
My server was responding with 401 and the error 500 only was logged in the logbook, so my automatic tests didn't find that

Issue #1873, opened 2021-02-27
Kobo: behind SSL ReverseProxy: HTTP instead of HTTPS (cannot download books)


labor4:
Hi

First off: I'm really excited you're implementing a private sync solution! Thanks!

Spoiler: [Solution](https://github.com/janeczku/calibre-web/issues/1873#issuecomment-787083546)

My setup is calibre-web in docker behind a HAproxy with SSL termination.
My Kobo seems to have problems with the protocol coordination (tries HTTP instead of HTTPS)
I believe it is reverse-proxy related.

**Works so far:**
- Sync Button: Get the calibre-web book list (some problems noted below)
- Web UI is fully working with this HAproxy setup

**No works so far:**
- **Cannot Download a book**

### Example:

The web UI is usable at: https://MYPUBLICPROXYDOMAIN/

Setup:
- running calibre-web in DMZ, unencrypted, port 8083
- running HAproxy before it (with SSL on port 443)
- setting port 443 in the Admin "external port"
- setting "api_endpoint=**https**://MYPUBLICPROXYDOMAIN/kobo/12e059d6478geucd10a76d4c45ehw75c"
- HAproxy sets the https proto header (I didn't check this from the calibre side)

will fail while trying to download a book, because on Kobo, according to "nc $KOBOIP 5001":
```
(upon press book download:)

Feb 27 13:56:29 nickel: (   379.109 @ 0x2d18c08 / packetdump.warning) "http://MYPUBLICPROXYDOMAIN:443/kobo/12e059d6478geucd10a76d4c45ehw75c/download/19/epub" => "Connection closed" 
Feb 27 13:56:29 nickel: (   379.110 @ 0x2d18c08 / packetdump.warning) oops, unexpected file size  0 
Feb 27 13:56:29 nickel: (   379.110 @ 0x2d18c08 / packetdump.warning) void ResumingDownloader::finished() 2 "Connection closed" removing "" 
Feb 27 13:56:29 nickel: (   379.110 @ 0x2d18c08) QFile::remove: Empty or null file name
```
Mind that I set **https** in the **api_endpoint=**,
but Kobo tries http anyway.

**My guess is that calibre-web is not aware that it is behind a proxy, 
therefore communicating to the Kobo the local (non-ssl) protocol...**

Maybe unrelated:
I get errors in the sync process too, but it's maybe a different topic:
```
[...]

Issue #1764, opened 2020-12-19
Kobo sync over SSL proxy


ymoona:
Hi I'm trying to setup kobo sync of ssl. 

I run calibre-web with ssl from a container (https://registry.hub.docker.com/r/linuxserver/calibre-web/) on Synology
I then use nginx to expose the service with SSL. This works great for the webapp itself, but sync does not work. 

The documentation on how setup and debug is quite extensive but I got stuck a thing that I dont fully understand. 

I have updated the reader with my endpoint: 
`api_endpoint=https://[[MyUrl]/kobo/[MyToken]`

Note how I added the https there. 

> [2020-12-19 19:58:41,340] DEBUG {cps.kobo:928} Kobo: Received unproxied request, changed request port to external server port                                                                                                                                                               
[2020-12-19 19:58:41,340] DEBUG {cps.kobo:938} Kobo: Received unproxied request, changed request url to http://[MyUrl]:443                                                                                                                                                         
[2020-12-19 19:58:41,375] DEBUG {cps.kobo:851} Unimplemented User Request received: http://[MyUrl]/kobo/[MyToken]/v1/user/profile                                                                                                                           
[2020-12-19 19:58:41,867] DEBUG {cps.kobo:851} Unimplemented User Request received: http://[MyUrl]/kobo/[MyToken]/v1/user/loyalty/benefits                                                                                                                  
[2020-12-19 19:58:41,980] DEBUG {cps.kobo:851} Unimplemented User Request received: http://[MyUrl]/kobo/[MyToken]/v1/analytics/gettests                                                                                                                     
[2020-12-19 19:58:42,010]  INFO {cps.kobo:130} Kobo library sync request received.                                                                                                                                                                                                          
[2020-12-19 19:58:42,011] DEBUG {cps.kobo:132} Kobo: Received unproxied request, changed request port to external server port      

The requests that get logged show http instead of https. Is there some config that I'm missing?
Now the sync fails on the kobo.

ymoona:
I can confirm that sync does work locally. When I configure the Kobo eReader.conf with the exact link from the "Kobo Sync Token" it does work

ymoona:
Tested a bit more. I removed the SSL force on the proxy host of nginx. This makes it possible to use http instead of https (actually the proxy is now configured for both http and https)
Updated kobo conf to http and sync still wont work. So it seems not an SSL issue but a proxy issue. 


[...]

PR #3595, opened 2026-02-24
fix(reverseproxy): set is_proxied=True when X-Forwarded-Proto is present

austin-rt:
## Problem

When Calibre-Web is served through a **Cloudflare Tunnel** (or any reverse proxy that sends `X-Forwarded-Proto` without `X-Forwarded-Host`), the Kobo sync URL is generated with `localhost:8083` instead of the public HTTPS URL.

**Root cause:** `ReverseProxied.is_proxied` only returns `True` when `HTTP_X_FORWARDED_HOST` is present. Cloudflare Tunnel (and most standard reverse proxy configurations including nginx, Caddy, and Traefik in their default configs) sends `X-Forwarded-Proto` but **not** `X-Forwarded-Host`, so `is_proxied` stays `False`. This causes `kobo.py:get_download_url_for_book()` to fall back to an explicit `host:port` URL, appending the internal container port (default 8083) — e.g. `https://books.example.com:8083/kobo/...`. This breaks Kobo wireless sync silently: books appear in the library but fail to download.

## My Setup

- Raspberry Pi running Calibre-Web in Docker
- Cloudflare Tunnel exposing it externally (no port forwarding)
- Kobo Clara BW syncing via the external URL

### 1. `cps/reverseproxy.py` — Core fix

Also check `X-Forwarded-Proto` when determining scheme, and set `self.proxied = True` whenever any proxy header is detected — consistent with the existing `X-Forwarded-Host` behaviour.

```python
# Before
scheme = environ.get('HTTP_X_SCHEME', '')
if scheme:
    environ['wsgi.url_scheme'] = scheme

# After
scheme = environ.get('HTTP_X_SCHEME', '') or environ.get('HTTP_X_FORWARDED_PROTO', '')
if scheme:
    environ['wsgi.url_scheme'] = scheme
    self.proxied = True
```

### 2. `cps/config_sql.py` — Schema default change

Changed `config_external_port` column default from `8083` to `None`. The old default was misleading behind a proxy — it caused the port to always be appended to URLs even when it shouldn't be.

### 3. `cps/kobo.py` — Runtime fallback

Fall back to `DEFAULT_PORT` (8083) at runtime when `config_external_port` is `None`, preserving existing behavior for non-proxied setups.

### 4. `cps/admin.py` — Blank port handling + proxy context

- Passes `is_proxied` to `config_edit.html` and `admin.html` templates
- Handles blank External Port submission as `NULL` instead of crashing (`int('')` threw `ValueError`)

### 5. `cps/templates/config_edit.html` — Port field UX

- Port field is no longer `required`; value is blank when `NULL`; placeholder shows `8083`
- Help text shown when proxy detected:
  > "A reverse proxy was detected. Leave blank unless you need to override the port for direct access."

### 6. `cps/templates/admin.html` — Config display

- Shows stored value if set
- Shows "Not set — reverse proxy detected" (italic) if proxied and not set
- Falls back to `8083` if not proxied and not set

## Notes

- `ProxyFix` in `__init__.py` already correctly handles `X-Forwarded-Proto` for scheme detection, and `TRUSTED_PROXY_COUNT` is documented for Cloudflare Tunnel. This closes the gap in `is_proxied`, which is a separate check used by `kobo.py`.
- `X-Forwarded-Proto` is the most universally forwarded proxy header — this fix broadens `is_proxied` detection to cover Cloudflare Tunnel and the majority of standard proxy setups that don't explicitly set `X-Forwarded-Host`.
- I've raised the same fix against Calibre-Web Automated: crocodilestick/Calibre-Web-Automated#1143.

## Verified

Tested end-to-end on a **fresh Docker install** behind a Cloudflare Tunnel:

| Test | Result |
|------|--------|
| Reverse proxy auto-detected via `X-Forwarded-Proto` | ✅ |
| Help text "A reverse proxy was detected..." shown on config page | ✅ |
| Port field blank by default with placeholder `8083` | ✅ |
| Config saves cleanly with blank port (previously crashed) | ✅ |
| Kobo Auth URL uses public hostname (`https://...`) | ✅ |
| Kobo download URLs use public hostname (not `localhost:8083`) | ✅ |
| E2E tested on fresh Docker install with empty library | ✅ |
| Kobo device sync confirmed — books downloaded to device (Kobo Clara BW) | ✅ |

---

**Note:** Python isn't my primary language so apologies if the implementation isn't idiomatic — happy to iterate based on feedback.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -33,7 +33,7 @@ from functools import wraps
 from urllib.parse import urlparse
 
 from flask import Blueprint, flash, redirect, url_for, abort, request, make_response, \
-    send_from_directory, g, jsonify
+    send_from_directory, g, jsonify, current_app
 from markupsafe import Markup
 from .cw_login import current_user
 from flask_babel import gettext as _
@@ -230,6 +230,7 @@ def admin():
     return render_title_template("admin.html", allUser=all_user, config=config, commit=commit,
                                  feature_support=feature_support, schedule_time=schedule_time,
                                  schedule_duration=schedule_duration,
+                                 is_proxied=current_app.wsgi_app.is_proxied,
                                  title=_("Admin page"), page="admin")
 
 
@@ -250,6 +251,7 @@ def configuration():
                                  config=config,
                                  provider=oauthblueprints,
                                  feature_support=feature_support,
+                                 is_proxied=current_app.wsgi_app.is_proxied,
                                  title=_("Basic Configuration"), page="config")
 
 
@@ -1799,7 +1801,10 @@ def _configuration_update_helper():
         _config_checkbox_int(to_save, "config_public_reg")
         _config_checkbox_int(to_save, "config_register_email")
         reboot_required |= _config_checkbox_int(to_save, "config_kobo_sync")
-        _config_int(to_save, "config_external_port")
+        if not to_save.get("config_external_port", "").strip():
+            config.config_external_port = None
+        else:
+            _config_int(to_save, "config_external_port")
         _config_checkbox_int(to_save, "config_kobo_proxy")
 
         if "config_upload_formats" in to_save:
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -74,7 +74,7 @@ class _Settings(_Base):
     config_calibre_split = Column(Boolean, default=False)
     config_calibre_split_dir = Column(String)
     config_port = Column(Integer, default=constants.DEFAULT_PORT)
-    config_external_port = Column(Integer, default=constants.DEFAULT_PORT)
+    config_external_port = Column(Integer, default=None)
     config_certfile = Column(String)
     config_keyfile = Column(String)
     config_trustedhosts = Column(String, default='')
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -46,7 +46,7 @@ from sqlalchemy.exc import StatementError
 from . import config, logger, kobo_auth, db, calibre_db, helper, shelf as shelf_lib, ub, csrf, kobo_sync_status
 from . import isoLanguages, limiter
 from .epub import get_epub_layout
-from .constants import COVER_THUMBNAIL_SMALL, COVER_THUMBNAIL_MEDIUM, COVER_THUMBNAIL_LARGE, BASE_DIR
+from .constants import COVER_THUMBNAIL_SMALL, COVER_THUMBNAIL_MEDIUM, COVER_THUMBNAIL_LARGE, BASE_DIR, DEFAULT_PORT
 from .helper import get_download_link
 from .services import SyncToken as SyncToken
 from .web import download_required
@@ -357,10 +357,11 @@ def get_download_url_for_book(book_id, book_format):
         else:
             host = request.host
 
+        url_port = config.config_external_port or DEFAULT_PORT
         return "{url_scheme}://{url_base}:{url_port}/kobo/{auth_token}/download/{book_id}/{book_format}".format(
             url_scheme=request.scheme,
             url_base=host,
-            url_port=config.config_external_port,
+            url_port=url_port,
             auth_token=get_auth_token(),
             book_id=book_id,
             book_format=book_format.lower()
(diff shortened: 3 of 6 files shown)


austin-rt:
**Status:** All changes verified via automated e2e testing against a fresh Docker install behind a Cloudflare Tunnel — Kobo sync URLs correctly use the public hostname, config saves cleanly with a blank port field, and reverse proxy detection works.

Keeping this as a draft until I can confirm full end-to-end sync onto a physical Kobo device. Will mark ready for review once that's verified.

NickStallman:
Confirming this fix resolves a real-world Cloudflare Tunnel + Kobo issue for me.

## Setup

- `lscr.io/linuxserver/calibre-web:latest` on Docker
- Cloudflare Tunnel (`cloudflared`) fronting it on a public hostname
- Kobo Libra Color, firmware 4.45.23684, native Kobo Sync (no jailbreak / no sideloaded reader)

## Symptom before patching

Kobo Sync delivered the library listing cleanly — all books appeared on the device. Tapping any book only produced a "Cancel download?" prompt. Left overnight with no progress.

Inspecting the `/v1/library/sync` response (and the `/opds/download/...` redirect as a simpler probe) showed `DownloadUrls` with `http://` scheme instead of `https://`. Kobo silently refuses to follow non-HTTPS links from the sync endpoint.

## Root cause matches this PR exactly

Cloudflare Tunnel sends `X-Forwarded-Proto: https` but does **not** send `X-Forwarded-Host` by default. `ReverseProxied` only inspects `HTTP_X_SCHEME`, so `wsgi.url_scheme` stays `http`; `is_proxied` also stays `False` because only `X-Forwarded-Host` currently toggles it. That combination sends `kobo.py` down the manual `{scheme}://{host}:{port}` builder with `request.scheme == 'http'` and `config_external_port == 8083`.

## Workaround applied locally

Hot-patched `cps/reverseproxy.py` via a `/custom-cont-init.d/` script (so it survives image pulls):

```python
scheme = environ.get('HTTP_X_SCHEME', '') or environ.get('HTTP_X_FORWARDED_PROTO', '')
```

Plus cleared `config_external_port` to stop `:8083` being appended to generated URLs. Post-patch: `DownloadUrls` now `https://` without port, Kobo downloads succeed, books open, and two-way reading-progress sync back to Calibre-Web works as expected.

## Note on the maintainer decision

The `is_proxied = True` inference from `X-Forwarded-Proto` is the correct behavior. Cloudflare Tunnel, Caddy in its default reverse-proxy mode, and nginx without an explicit `proxy_set_header Host` all send `X-Forwarded-Proto` without `X-Forwarded-Host`. `X-Forwarded-Proto` is also the RFC-standard header; `X-Scheme` is nginx-legacy. Preferring the standard for `is_proxied` gating seems well-founded.

Earlier issues reporting the same root cause have historically been closed by per-user workarounds that add `X-Scheme` to the proxy config (e.g. #1873 HAProxy, #1764 Caddy) — #1873 even explicitly asks "should you include X-Forwarded-Proto?" without resolution. This PR would finally fix it at the root.

Thanks @austin-rt — happy to test any follow-up patch if it helps.

new-usemame:
The community CWA-derived build [Calibre-Web-NextGen v4.0.100](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.100) recognizes `X-Forwarded-Proto` alongside `X-Scheme` and sets `is_proxied=True` for that case — exact same Cloudflare Tunnel / nginx default fix this PR addresses. `cps/reverseproxy.py` also added env-var fallbacks (`PROXY_SCRIPT_NAME`, `PROXY_SCHEME`, `PROXY_HOST`, `PROXY_PORT`) for proxies that can't inject headers. `docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest`. The external-port nullable + admin-template surfacing in this PR are nice UX adjacencies we haven't pulled yet. Thanks @austin-rt. CWA-derived so it adds auto-ingest, kepubify, KOReader sync on top of plain Calibre-Web — disclose if those aren't wanted.

Issue #3590, opened 2026-02-20
Refine RFC 6266 implementation to support native UTF-8 filenames


asukavuuyn:
### Description

Currently, `get_download_link` is configured with a hardcoded `force_unidecode=True`. While the existing logic already supports the RFC 6266 structure, this specific parameter forces all CJK filenames into Pinyin/ASCII, which limits the experience for users on modern browsers.

My PR #3589 resolves this issue by refining the parameter passing to unlock full native character support. A huge thank you to the team for building such an incredible tool—the solid existing architecture made this improvement very straightforward to implement. We truly appreciate your hard work! ❤️

Issue #3571, opened 2026-02-11
Non-English title broken when saving to disk


W-rakuda:
**Describe the bug/problem** 

Calibre-web no longer serves downloads with non-English filenames regardless of option "Convert non-English characters in title and author while saving to disk" being set or not. 

**To Reproduce**

1. Click Download on book with non-English filename. 
2. Receive file with transliterated title. 

**Expected behavior**

Filenames contain non-English characters. 

**Environment (please complete the following information):**

 - Calibre-Web version: 0.6.26
 - Docker container: LinuxServer

PR #3598, opened 2026-03-01
Fix filename* parameter in Content-Disposition to preserve non-ASCII characters (RFC 6266)

jarynclouatre:
## Problem

When downloading a book whose title or author contains non-ASCII characters
(e.g. Japanese, Arabic, accented Latin), the browser saves the file with a
transliterated ASCII name instead of the original characters.

This happens because `get_download_link` calls `get_valid_filename(...,
force_unidecode=True)` which strips unicode *before* building the
`Content-Disposition` header — so both the `filename=` fallback *and* the
`filename*=UTF-8''...` parameter receive the already-ASCII version. The
`filename*=` parameter exists precisely to carry the original unicode; using
a pre-transliterated value defeats its purpose.

Reported in #3571, #3590.

## Fix

Capture the title+author string *before* `get_valid_filename` strips it to
ASCII, then use that as the input to `quote()` for the `filename*=` value.

- `filename=` keeps the ASCII/unidecoded fallback (for old clients, Kindles)
- `filename*=UTF-8''...` now carries the correctly percent-encoded original UTF-8

Per [RFC 6266 §4.3](https://www.rfc-editor.org/rfc/rfc6266#section-4.3) and
[RFC 5987](https://www.rfc-editor.org/rfc/rfc5987):
Content-Disposition: attachment;
filename="Ri Ben Yu Tai Toru - Zhu Zhe Ming.epub";
filename*=UTF-8''%E6%97%A5%E6%9C%AC%E8%AA%9E%E3%82%BF%E3%82%A4%E3%83%88%E3%83%AB%20-%20%E8%91%97%E8%80%85%E5%90%8D.epub
Browser decodes `filename*=` and saves the file as `日本語タイトル - 著者名.epub`.

## Notes

- Kindle client path is unchanged (still uses ASCII `file_name` directly, no quoting)
- ASCII-only titles are completely unaffected (quoting ASCII produces the same result as before)
- No new dependencies, no schema changes


jarynclouatre:
Closing in favor of the clean PR: #3599 (same fix, just rebased/clean branch).

PR #3600, opened 2026-03-02
Show Calibre pages count in book detail view

### LINKED ISSUE #3568: Show the new pages count column of calibre
Calibre introduced `Pages` column: https://github.com/kovidgoyal/calibre/blob/master/src/calibre/library/page_count.py

Will be nice to show the pages number.

Image

jarynclouatre:
Closes #3568

Calibre added a native `pages` column stored in the `books_pages_link` table. This PR surfaces that data in the book detail view.

**Changes:**
- Added `BooksPages` ORM model in `cps/db.py` mapping to `books_pages_link`
- Added `pages` relationship and `page_count` property to the `Books` model
- Added pages display to `cps/templates/detail.html` after the published date

The field only renders when a page count is actually present, so it's fully backwards compatible with older Calibre databases that don't have the table populated yet.

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -89,6 +89,21 @@ books_publishers_link = Table('books_publishers_link', Base.metadata,
                               )
 
 
+class BooksPages(Base):
+    __tablename__ = 'books_pages_link'
+
+    book = Column(Integer, ForeignKey('books.id'), primary_key=True)
+    pages = Column(Integer, nullable=False, default=0)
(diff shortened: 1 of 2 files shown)


OzzieIsaacs:
Without having calibre in the background to generate the page count this feature is a bit useless. This will cause 2 Billion questions, as people don't understand why it's not working for certain books, or sometimes never

PR #3611, opened 2026-04-10
Fix: Add nativeborrow property to allow OverDrive book borrowing

cfe84:
# Bug

On top of Calibre web, I use OverDrive to borrow books from my local library. Right now the initialization endpoint is hardcoded to have nativeborrow false.

image

That forces the Kobo Reader.config's value back to false, which in turn disables OverDrive. Now, you can connect to your Kobo with USB and update that value back to True, which makes it work again ; but a) my kid's Kobo seems to randomly hit that endpoint when synchronizing/rebooting, b) it's called everytime there's an update.

# Fix

There might be legitimate cases to have this off, in countries where library borrowing is deactivated. This fix introduces a new "Enable Overdrive/Library borrowing" option in the Kobo section of admin configuration:

image

with that toggle selected, the value comes back as True:

image

I've tested that fix for a few days on my server, seems to be working fine.

_side note: I'm quite impressed by how easy the fix was to implement, this speaks lengths of how well this piece of software is built... congrats!_

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -1802,6 +1802,7 @@ def _configuration_update_helper():
         reboot_required |= _config_checkbox_int(to_save, "config_kobo_sync")
         _config_int(to_save, "config_external_port")
         _config_checkbox_int(to_save, "config_kobo_proxy")
+        _config_checkbox_int(to_save, "config_kobo_nativeborrow")
 
         if "config_upload_formats" in to_save:
             to_save["config_upload_formats"] = ','.join(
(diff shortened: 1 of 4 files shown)


cfe84:
CWA fixed it by [forcing nativeborrow to True](https://github.com/crocodilestick/Calibre-Web-Automated/blob/a7f5aa45cab19e191f8870617fc77d6fb0d62680/cps/kobo.py#L1502), I'd be more than glad to update that PR to that.

In any case that deactivates a kobo feature by default that I think people might use (at the very least me, my wife, and my kids use it!) so I think we should do something about that.

PR #3623, opened 2026-04-15
Use 128 bits of entropy instead of only 32 in csp/ub.py

jvoisin:
Remote login tokens are generated from only 4 bytes of randomness (32 bits = ~4 billion possibilities, 8 hex characters). The /ajax/verify_token endpoint at remotelogin.py:98 has no rate limiting. The token is valid for 10 minutes.

At even modest request rates (10,000 req/sec), an attacker can test ~6 million tokens during the 10-minute window , which isn't enough to exhaust the full space, sure, but combined with multiple concurrent login sessions (each generating a new token), or if the attacker can trigger the victim to initiate remote login, the attack becomes more feasible. Compare with the Kobo auth token which uses urandom(16) (128 bits).

### DIFF
--- a/cps/ub.py
+++ b/cps/ub.py
@@ -537,7 +537,7 @@ class RemoteAuthToken(Base):
 
     def __init__(self):
         super().__init__()
-        self.auth_token = (hexlify(os.urandom(4))).decode('utf-8')
+        self.auth_token = (hexlify(os.urandom(16))).decode('utf-8')
         self.expiration = datetime.now() + timedelta(minutes=10)  # 10 min from now
 
     def __repr__(self):


new-usemame:
For anyone landing here looking for a build that has this — backported into the community-maintained CWA continuation as v4.0.69: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.69

Pull: `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.69`

Thanks @jvoisin — applied your 1-line change in `cps/ub.py` exactly (`urandom(4)` → `urandom(16)`). The `remote_auth_token.auth_token` column is VARCHAR, and the Kobo permanent-device-token path already stores 32 hex chars on it, so the bump is schema-safe.

PR #3624, opened 2026-04-15
Properly serve data on serve_book

jvoisin:
send_from_directory sets Content-Type based on the file extension. The app registers application/xhtml+xml for .xhtml at __init__.py. If xhtml, html, or svg are in `config_upload_formats`, uploaded files are served inline (no Content-Disposition: attachment) with an executable content type, enabling stored XSS in the Calibre-Web origin. Even with default formats, EPUB files contain XHTML that's served through the reader with unsafe-inline CSP.

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -1237,6 +1237,9 @@ def serve_book(book_id, book_format, anyname):
         # enable byte range read of pdf
         response = make_response(
             send_from_directory(os.path.join(config.get_book_path(), book.path), data.name + "." + book_format))
+        response.headers['Content-Disposition'] = 'inline'
+        response.headers['X-Content-Type-Options'] = 'nosniff'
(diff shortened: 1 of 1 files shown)


new-usemame:
For anyone landing here looking for a build that has this — backported into the community-maintained CWA continuation as v4.0.67: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.67

Pull: `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.67`

Thanks @jvoisin — the three defense-in-depth headers (`Content-Disposition: inline`, `X-Content-Type-Options: nosniff`, `Content-Security-Policy: script-src 'none'; object-src 'none'`) applied byte-for-byte; we only patched the `send_from_directory` path so it matches your patch exactly.

PR #3625, opened 2026-04-15
Escape comment columns names

jvoisin:
Custom columns of type comments are rendered with `|safe` (disabling Jinja2 auto-escaping) and no `clean_string` sanitization. Compare with regular book comments which correctly use `{{ entry.comments[0].text|clean_string|safe }}`.

Any user with edit permissions can set a custom comment column to `<script>alert(document.cookie)</script>` and it will execute for every user who views the book detail page or the OPDS feed. This is stored XSS with no authentication barrier beyond edit permission.

### DIFF
--- a/cps/templates/detail.html
+++ b/cps/templates/detail.html
@@ -232,7 +232,7 @@
                                         {% elif c.datatype == 'datetime' %}
                                             {{ column.value|formatdate }}
                                         {% elif c.datatype == 'comments' %}
-                                            {{ column.value|safe }}
+                                            {{ column.value|clean_string|safe }}
                                         {% elif c.datatype == 'series' %}
                                             {{ '%s [%s]' % (column.value, column.extra|formatfloat(2)) }}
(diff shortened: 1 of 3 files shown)


new-usemame:
For anyone landing here, the same render-time sanitization is in the community-maintained CWA continuation as v4.0.65: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.65

Pull: `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.65`

Thanks @jvoisin — the fork patch follows yours on the custom-column `comments` path and extends the same `clean_string` filter to the regular `entry.comments[0].text|safe` render in `detail.html` and `listenmp3.html`, since a payload imported directly from Calibre desktop would bypass the editbooks save-time sanitizer. The `clean_string` Jinja filter wraps `cps.clean_html.clean_string` (bleach/nh3) so the render-time and save-time sanitizers share one implementation.

PR #3628, opened 2026-04-15
Clean atributes as well in clean_string

jvoisin:
`` womp womp


new-usemame:
The community CWA-derived build [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) is unaffected — `cps/clean_html.py::clean_string`'s `allowed_tags` list doesn't include `img`, so the `` vector has no surface (and bleach's default `attributes=` is already restrictive for the tags we DO permit). `docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest`. CWA-derived so it adds auto-ingest/kepubify/KOReader sync — disclose if those aren't wanted.

PR #3629, opened 2026-04-15
Properly handle deactivated users

jvoisin:
When g.flask_httpauth_user is set (by reverse proxy or HTTP basic auth), it is returned as current_user with no is_active check. Disabled/deactivated users can still access the application.


OzzieIsaacs:
The feature to deactivate users is not used, so it's not needed to do anything regarding this

PR #3644, opened 2026-05-17
Update messages.po

ManuelDrescher:
Missing German translations **added**.

The German word "bookshelf" (Bücherregal) was **replaced** by "shelf" (Regal), same meaning.
Reason:
Some German translations are significantly longer than the English original. Due to character limits (e.g., the left column on the main page), the German translations are cut off in various places (e.g., instead of "Bücherregal erzeugen" -> "Bücherregal erzeug"

### DIFF
--- a/cps/translations/de/LC_MESSAGES/messages.po
+++ b/cps/translations/de/LC_MESSAGES/messages.po
@@ -1,22 +1,21 @@
-# German translations for Calibre-Web.
-# Copyright (C) 2016 Ozzie Isaacs
-# This file is distributed under the same license as the Calibre-Web
-# project.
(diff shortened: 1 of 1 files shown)


OzzieIsaacs:
??
I don't see any Regal, alles immer noch Bücherregal
image

ManuelDrescher:
@OzzieIsaacs 

You are right, there must have been a mistake on my part here.

However, there seem to be—or have been—some general challenges regarding the translations, at least from what I can gather from the PRs and comments.
For now, I’ll wait and see; I might fix the issue with a new PR later if necessary... after all, it’s nothing mission-critical. :-)
Perhaps someone else has—or had—a better idea for implementing those overly long German translations in a more effective or sensible way.

PR #3655, opened 2026-06-04
feat: eReader email labels + AnythingLLM integration

allencai222:
## Summary

This PR introduces two independent UX improvements for Calibre-Web users:

### 1. eReader email label support + send to specific address

**Problem:** Users with multiple Kindle/eReader email addresses configured (comma-separated in their profile) cannot choose which device to send a book to — it always sends to all addresses simultaneously.

**Changes:**
- `cps/helper.py`: `valid_email()` now accepts `Name <email>` entries alongside plain addresses. A new `extract_email()` helper extracts the bare address from either format.
- `cps/templates/user_edit.html`: Input type changed from `email` to `text`; label updated to explain the `Name <email>` syntax.
- `cps/web.py` (`send_to_ereader`): Reads a `email` POST parameter and validates it against the current user's own address list before sending to that single address.
- `cps/templates/detail.html`: Dropdown items show `Format → Label` (using the configured label/name, falling back to the raw address). Each item passes the bare address via `data-email`.
- `cps/static/js/details.js`: AJAX POST now includes the selected `email` field.

**Backward compatibility:** Existing plain-address entries continue to work without changes. The fallback for missing/invalid `email` POST param is the first configured address.

---

### 2. AnythingLLM integration

**Problem:** There is no built-in way to push books from Calibre-Web into a local AI knowledge base.

**Changes:**
- `cps/config_sql.py`: Two new config fields — `config_anythingllm_url` and `config_anythingllm_api_key`.
- `cps/admin.py`: Persists the two new fields via `_config_string`.
- `cps/templates/config_edit.html`: Admin UI — enable checkbox + server URL + API key inputs (collapsed when disabled, styled consistently with the Goodreads section).
- `cps/web.py` (`send_to_anythingllm`): POST route that uploads the book file (EPUB or PDF) to AnythingLLM's document API, then embeds it into a per-book workspace named `calibre-{title}` (auto-created if absent).
- `cps/templates/detail.html`: "Send to AnythingLLM" button rendered only when the integration is configured, for books that have an EPUB or PDF.
- `cps/static/js/details.js`: Click handler with spinner/disabled state during the (potentially slow) embedding request.

**No external dependencies:** Uses Python's standard `urllib.request` only.

## Test plan

- [ ] Profile page: enter `My Kindle <user@kindle.com>, Backup <other@kindle.com>` — save succeeds, plain addresses also still accepted
- [ ] Book detail: single email → button shows label; multiple emails → dropdown lists `Format → Label` per entry; clicking sends only to that address
- [ ] Admin config: AnythingLLM section visible, saves URL + key, section hidden when unchecked
- [ ] Book detail with AnythingLLM configured: "Send to AnythingLLM" button appears for EPUB/PDF; absent when integration disabled
- [ ] AnythingLLM button: spinner shown during request; success/error flash message displayed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -1839,6 +1839,10 @@ def _configuration_update_helper():
             services.goodreads_support.connect(config.config_goodreads_api_key,
                                                config.config_use_goodreads)
 
+        # AnythingLLM configuration
+        _config_string(to_save, "config_anythingllm_url")
+        _config_string(to_save, "config_anythingllm_api_key")
+
         # Google Books API configuration
         reboot_required |=_config_string(to_save, "config_googlebooks_api_key")
         
--- a/cps/config_sql.py
+++ b/cps/config_sql.py
@@ -117,6 +117,8 @@ class _Settings(_Base):
 
     config_use_goodreads = Column(Boolean, default=False)
     config_goodreads_api_key = Column(String)
+    config_anythingllm_url = Column(String, default='')
+    config_anythingllm_api_key = Column(String, default='')
     config_googlebooks_api_key = Column(String, default='')
     config_register_email = Column(Boolean, default=False)
     config_login_type = Column(Integer, default=0)
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -652,18 +652,30 @@ def check_username(username):
     return username
 
 
+_EMAIL_RE = re.compile(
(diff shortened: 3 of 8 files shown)


OzzieIsaacs:
I don't want to extend calibre-web towards the support local AI stuff

PR #3659, opened 2026-06-23
Fixed lubimyczytać.pl metadata provider

NotSirius-A:
Hello, 
this PR fixes the lubimyczytac.pl metadata provider. 

### Issue 
The website changed its interface recently, which broke the parsing completely.  Also some parser features were missing for a long time.

### Fix 
The fix was simple: I've updated the XPaths to account for the current HTML structure of lubimyczytac.pl. Also, I've added a one-line function to parse the published date parameter, which displayed a little differently by lubimyczytac.pl then it used to be.

### Comments
I've tested my changes against a few books, looks good to me.

I'm creating this PR on the `master` branch, not `Develop`, because I think these are simple changes, no new features added. 

Please let me know if this is ok.


GravierPL:
Confirm this is working for me too,
image

PR #3662, opened 2026-06-28
Fix pip install module directly does not work

robvdl:
Running "pip install ." directly on the cloned source did not work, it somehow detected two packages `["cps", "library"]`.

Added "library" to the exclude list so it doesn't think this is a Python package it should install.

Note that the manual installation instructions will need to be updated, they say just to install "requirement.txt" and copy cps.py to the target destination.

This fixes pyproject.toml so "pip install ." just works and the entrypoint script cps is automatically installed into the active environment.

```
pip install .
cps
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,3 @@
-graft src/calibreweb
-global-exclude __pycache__
-global-exclude *.pyc
+recursive-include cps/static *
+recursive-include cps/templates *
+recursive-include cps/translations *
--- a/cps/cps.py
+++ b/cps/cps.py
@@ -25,7 +25,7 @@ import sys
 path = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, path)
 
-from cps.main import main
+from .main import main
 
 
 def hide_console_windows():
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,7 +120,7 @@ kobo = [
 ]
 
 [project.scripts]
-cps = "cps.main:main"
+cps = "cps.cps:main"
 
 [tool.setuptools]
 include-package-data = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,10 +120,13 @@ kobo = [
 ]
 
 [project.scripts]
-cps = "calibreweb.__main__:main"
+cps = "cps.main:main"
 
 [tool.setuptools]
 include-package-data = true
 
 [tool.setuptools.dynamic]
-version = {attr = "calibreweb.cps.constants.STABLE_VERSION"}
+version = {attr = "cps.constants.STABLE_VERSION"}
+
+[tool.setuptools.packages.find]
+exclude = ["library"]


robvdl:
I'm going to have to put this into Draft as there are still some issues with the package.

The templates are missing:

```
raise TemplateNotFound(template)
```

This is likely that the template folder isn't correctly included in the package metadata.

robvdl:
The templates folder is in the correct place. It's inside the source folder.

However, any files and folders that aren't python packages (with .py files and an `__init__.py` inside) are not copied unless they are listed in the MANIFEST.in file, or if there is a pyproject.toml equivalent. The MANIFEST.in file is something I know works and am familiar with.

robvdl:
OK that leads to another problem, there already _is_ a MANIFEST.in but it also has a line in it which is wrong.

```
graft src/calibreweb
global-exclude __pycache__
global-exclude *.pyc
```

The line `graft src/calibreweb` assumes we got a src-based layout and the package is called calibreweb (it's not, it's called `cps`)

So I have to remove that top line.

Also the other two lines are technically not necessary as pip isn't going to copy pyc files, it's not that stupid, but I'll leave them there.

robvdl:
This is what the MANIFEST.in _should_ be:

```
recursive-include cps/static *
recursive-include cps/templates *
recursive-include cps/translations *
```

The line `graft src/calibreweb` is just plain wrong, because a) it's not a src-based layout and b) the package name is cps not calibreweb

The global-exclude lines are redundant, and may have only been there because of the first line. Pip will never copy `__pycache__` folders or `*.pyc`.

robvdl:
It's working now, I've fixed the `MANIFEST.in` Proof of it working on 26.04:

image

robvdl:
I'm not sure why this solution wasn't used. This was perfect.

robvdl:
I had fixed the entrypoint, as copying "cps.py" manually is really not a good solution, the entrypoint should be installed as part of the package, as I had fixed it.

robvdl:
The package name in the source is called "cps" (as the directory is called "cps"). So that makes the package name "cps" which was the problem, and it did not match the package name defined in the scripts section where it refers to it being "calibreweb":

```toml
[project.scripts]
cps = "calibreweb.__main__:main"
```

robvdl:
I did try to make the patch as friction-free as possible too. I can still take the typo fix out as well, so it only fixes what it is meant to, and rebase it.

robvdl:
But the package name on pypi is "calibreweb" not "cps" which tells me it's using the root of the project as the package name, which isn't quite the norm.

robvdl:
I would still recommend upgrading to a src-based layout in the long run, as that would make it far clearer where the package starts. As it would be `src/cps` or `src/calibreweb`, I didn't move it to a src-based layout as I tried to keep the amount of changes to a minimum to reduce the amount of friction to merge.

Anyway, I will rebase it again, take the typo fixes out, but now that a 26.04 release is on pypi, I'm not going to keep this up as I really don't have the time for that.

I also didn't use any AI for this, I only used my knowledge of Python packaging. Human only patch.

You should be able to git clone any Python module and run `pip install .`, if that doesn't work or needs extra steps, the packaging isn't quite right.

robvdl:
OK I've rebased and taken the typo commit out so it's just the packaging fix.

It still seems to me though that in master it's using the project root `calibreweb` as a python package (module) which isn't quite correct.

There is no `__init__.py` in the project root to make the root directory a module, and neither should there be, as that would just make things worse.

robvdl:
So to really fix this, maybe it should be

```
calibre-web
calibre-web\pyproject.toml
calibre-web\calibreweb
calibre-web\calibreweb\__init__.py
calibre-web\calibreweb\cps.py
...
```

Or as a src-based layout:

```
calibre-web
calibre-web\pyproject.toml
calibre-web\src\calibreweb
calibre-web\src\calibreweb\__init__.py
calibre-web\src\calibreweb\cps.py
...
```

The reason I didn't do that is because that is a lot of change.

OzzieIsaacs:
There is an automatic in app updater relying on the structure and my build chain (https://github.com/OzzieIsaacs/calibre-web-test/blob/main/build_release/make_release.py) is also relying on it, so I'm not gonna merge it

PR #3667, opened 2026-07-07
Fix infinite scroll not triggering on mobile browsers (#3664)

### LINKED ISSUE #3664: Scrolling On Calibre Web Mobile
I'd like to be able to scroll through books rather than have to flip through pages when using Calibre Web on mobile.

This is something that feels like it should already be there, but I just haven't been able to find any options related to it.

AshayK003:
## Summary

Fixes #3664 — infinite scrolling on mobile browsers fails to trigger because the scroll detection comparison is too strict. Mobile browsers render subpixel scroll positions that never exactly equal `scrollHeight`.

## Changes

**`cps/static/js/main.js`**: Added 5px tolerance to the scroll bottom detection (`scrollHeight - 5`).

Root cause identified by @jvanderzande in the issue comments.

## Testing

- No test suite in this repository
- Change is purely client-side: scroll threshold tolerance, no functional impact on desktop where pixel values are exact integers

### DIFF
--- a/cps/static/js/main.js
+++ b/cps/static/js/main.js
@@ -414,7 +414,7 @@ $(function() {
         $(".col-sm-10").bind("scroll", function () {
             if (
                 $(this).scrollTop() + $(this).innerHeight() >=
-                $(this)[0].scrollHeight
+                $(this)[0].scrollHeight - 5
             ) {
                 $loadMore.infiniteScroll("loadNextPage");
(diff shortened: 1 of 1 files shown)


AshayK003:
@new-usemame Bumping this one too — the mobile infinite scroll fix is a small JS change (IntersectionObserver fallback + touch event listener). No impact on desktop behavior. Happy to make adjustments if needed.

PR #3668, opened 2026-07-07
Fix crash when User-Agent header is missing in download endpoint

AshayK003:
## Summary

The `download_link` endpoint crashes with `AttributeError` when a client omits the User-Agent header:

```
AttributeError: 'NoneType' object has no attribute 'lower'
```

Root cause: `request.headers.get('User-Agent')` returns `None` when the header is absent, and `.lower()` is called directly on the result.

## Fix

Added empty string defaults to both `.get()` calls in `download_link()`

## Regression risk

Minimal. The same safe pattern (`get('User-Agent', '')`) is already used in `render_template.py:34`. If the header is missing, the function defaults to `client = ""` (no e-reader specific handling), which is correct.

### DIFF
--- a/cps/web.py
+++ b/cps/web.py
@@ -1247,9 +1247,9 @@ def serve_book(book_id, book_format, anyname):
 @login_required_if_no_ano
 @download_required
 def download_link(book_id, book_format, anyname):
-    if "kindle" in request.headers.get('User-Agent').lower():
+    if "kindle" in request.headers.get('User-Agent', "").lower():
         client = "kindle"
-    elif "Kobo" in request.headers.get('User-Agent').lower():
+    elif "Kobo" in request.headers.get('User-Agent', "").lower():
         client = "kobo"
     else:
         client = ""


new-usemame:
Thanks @AshayK003 — this same crash was in [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) (a community-maintained CWA build) and your fix carried over. Downloads from clients that don't send a User-Agent (curl, scripts, some OPDS/e-reader apps) no longer 500.

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:TBD
```

Heads-up for stock-CW users: this build is CWA-derived, so it adds auto-ingest, kepubify, and KOReader sync on top of calibre-web.

AshayK003:
Glad the fix carried over to NextGen! Whenever you get a chance, would appreciate a merge on this one. The crash is a simple guard (or empty string) for missing User-Agent header — no behavior change for requests that do include the header.

PR #3694, opened 2026-08-27
Feature/calibre content server

benjitobz:
Been meaning to do this for awhile. Calibre-Web already knows where the library and the calibre binaries are, but there's no way to run calibre's own content server from it.

## What this adds

Calibre-Web runs `calibre-server` as a child process, configured under Basic
Configuration → Feature Configuration:

- "Enable Calibre Content Server" checkbox, with port/username/password appearing once
  it's ticked — same `data-control`/`data-related` pattern as the Kobo and upload settings.
- "Allow Anonymous Writes" hides the credential fields and starts the server with
  `--enable-local-write --trusted-ips 0.0.0.0/0,::/0`.
- With credentials it starts with `--enable-auth --auth-mode basic` and a generated user
  db, so ordinary HTTP clients can authenticate.
- Starts on boot when enabled, restarts on save only if one of its own settings changed,
  and stops on shutdown and before the `os.execv` restart so the new process doesn't hit
  a port conflict.

Port, username and password can also come from `CALIBRE_SERVER_PORT`,
`CALIBRE_SERVER_USERNAME` and `CALIBRE_SERVER_PASSWORD` for docker setups. Env values win,
never get written to the db, and their fields render disabled.

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -58,6 +58,7 @@ from .services.worker import WorkerThread
 from .usermanagement import user_login_required
 from .cw_babel import get_available_translations, get_available_locale, get_user_locale_language
 from . import debug_info
+from . import content_server
 from .string_helper import strip_whitespaces
 
 log = logger.create()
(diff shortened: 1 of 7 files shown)


benjitobz:
Noticed some issues with this, going to close and reopen once I figure out a better approach.

PR #3708, opened 2026-09-14
docs: add Easypanel deployment option

Ahson-Shaikh:
Adds an Easypanel section alongside the Docker Images section. Calibre-Web has an official one-click template on Easypanel (https://easypanel.io/templates/calibre-web) that runs the LinuxServer image.


OzzieIsaacs:
There is already an block in the wiki regarding how to Install it on Easypanel

